PR #150, opened 2019-03-18
Add simple viewer example

williampMSFT:
Add a simple OpenGL-based viewer tool that viewer tool as an example so developers can quickly get started with their own viewer tool.

The simple viewer borrowed some code from the full K4AViewer tool; as part of this, cleaned up some of that code as well


wqs111000:
How to use this  simple viewer example? Can anyone add some detailed guidance?

Issue #151, opened 2019-03-18
Create a C based record sample


wes-b:
We currently have a CPP based record tool that is in the examples folder. It has gotten fairly complex, so we should consider moving the current example into the tools folder. We should also write a C-Based sample. C code is specifically needed because the header file is C based and we also need to test to confirm the C header stays compatible for C projects (right now all users of record are CPP). Over time we will make CPP, C# version of the simple sample for users of the SDK.

PR #246, opened 2019-04-16
Feature/support custom track record playback

### LINKED ISSUE #263: Support custom track record and playback

yijiew:
Closes #263 #151

 Add functionality to record and playback custom track data. Please find the API definitions in record.h and playback.h

-Record
1. Add custom track recording code to recorder.cpp
2. Expose the custom track recording API through record.h
3. Add API to add custom attachment to the recording
4. Add API to add tags that link to the custom tracks
5. Create a simple example to showcase how the API can be used.
6. Add unit test for the custom track recording APIs.

-Playback
1. Add feature to playback custom track and parse custom track information in playback.cpp
2. Expose the custom track playback API through playback.h
3. Add API to parse attachment from a recording
4. Define data_block handle to access to custom track timestamp and data
5. Add unit test for custom track playback


include/k4arecord/playback.h
```diff
@@ -363,6 +519,88 @@ K4ARECORD_EXPORT k4a_stream_result_t k4a_playback_get_next_imu_sample(k4a_playba
 K4ARECORD_EXPORT k4a_stream_result_t k4a_playback_get_previous_imu_sample(k4a_playback_t playback_handle,
                                                                           k4a_imu_sample_t *imu_sample);
 
+/** Read the next data block of a track with the given track name.
+ *
+ * \param playback_handle
+ * Handle obtained by k4a_playback_open().
+ *
+ * \param track_name
+ * The track name defines the track that you want to get the next data block from.
```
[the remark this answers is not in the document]
xthexder: The built-in tracks are now defined and documented `k4arecord/types.h`, Track names can also be discovered programatically with `k4a_playback_get_track_count()` and `k4a_playback_get_track_name()`

xthexder:
I only have 1 thing left to finish on this PR, and that is to update the custom track example.
I'll be finishing up changes to the custom track example shortly, so it would be good to get the final review of this PR started.

I've gone through and reviewed all the APIs / docs / tests, so those should be finalized apart from any additional feedback people have.

williampMSFT:
> /** \file playback.h

I think we just took the change to promote the C++ playback wrapper to the public API (see commit a281018825b728a81fd11da89f8fe36281ed1377) - we might want to update it to include the functions you've added here to maintain parity between the C and C++ headers

---
Refers to: include/k4arecord/playback.h:1 in 51628cd. [](commit_id = 51628cdc177e7bf38475d4f021fdd92efa6dc62e, deletion_comment = False)

include/k4arecord/playback.h
```diff
@@ -134,6 +134,203 @@ K4ARECORD_EXPORT k4a_result_t k4a_playback_get_calibration(k4a_playback_t playba
...
+/** Gets the name of a track at a specific index.
+ *
+ * \param playback_handle
+ * Handle obtained by k4a_playback_open().
+ *
+ * \param track_index
+ * The index of the track to read the name form.
+ *
+ * \param track_name
+ * Location to write the track name. This will be a UTF8 null terminated string. If a NULL buffer is specified, \p
```
williampMSFT: >\p [](start = 112, length = 2)

my doxygen-fu is weak so this might be fine, but it looks like the line wrap separated this annotation from the variable it refers to - if this actually works in doxygen it's fine, but I wouldn't be surprised if this doesn't work, might want to move to the next line

include/k4arecord/playback.h
```diff
@@ -134,6 +134,203 @@ K4ARECORD_EXPORT k4a_result_t k4a_playback_get_calibration(k4a_playback_t playba
...
+ * \endxmlonly
+ */
+K4ARECORD_EXPORT k4a_buffer_result_t k4a_playback_track_get_codec_id(k4a_playback_t playback_handle,
+                                                                     const char *track_name,
+                                                                     char *codec_id,
+                                                                     size_t *codec_id_size);
+
+/** Gets the codec context for a particular track.
+ *
+ * The codec context is codec-specific buffer that contains any required codec metadata that only known to the codec. It
```
williampMSFT: >The codec context is codec-specific buffer that contains any required codec metadata that only known to the codec. [](start = 3, length = 114)

nit: minor grammar fixes
... codec context is *a* codec-specific buffer ...
... metadata that is only known ...

xthexder:
> > /** \file playback.h
> 
> I think we just took the change to promote the C++ playback wrapper to the public API (see commit [a281018](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/commit/a281018825b728a81fd11da89f8fe36281ed1377)) - we might want to update it to include the functions you've added here to maintain parity between the C and C++ headers
> 
> Refers to: include/k4arecord/playback.h:1 in [51628cd](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/commit/51628cdc177e7bf38475d4f021fdd92efa6dc62e). [](commit_id = [51628cd](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/commit/51628cdc177e7bf38475d4f021fdd92efa6dc62e), deletion_comment = False)

Yeah, looks like that is in develop now. We should definitely have C/C++ parity for the 1.2.0 release, but maybe it would be best to split that up into a separate PR. This PR is already pretty big to review fully.

include/k4arecord/playback.h
xthexder: It looks like doxygen figures this out just fine ([See Here](https://microsoft.github.io/Azure-Kinect-Sensor-SDK/feature/support-custom-track-record-playback/group___functions_gaaa6e6d931691c5fde0db0a831286109f.html#gaaa6e6d931691c5fde0db0a831286109f)), but I'll change this anyway just so it's a bit more readable in the source code.

xthexder:
I've checked in all my changes, so this PR is ready for final review.

tests/RecordTests/UnitTest/playback_ut.cpp
```diff
@@ -93,7 +93,7 @@ TEST_F(playback_ut, open_large_file)
     k4a_stream_result_t stream_result = K4A_STREAM_RESULT_FAILED;
```
wes-b: I don't see any tests in this file that just validate inputs for the API's. Can you add that?

include/k4arecord/playback.h
```diff
@@ -370,6 +601,190 @@ K4ARECORD_EXPORT k4a_stream_result_t k4a_playback_get_next_imu_sample(k4a_playba
...
+ *
+ * \xmlonly
+ * <requirements>
+ *   <requirement name="Header">playback.h (include k4arecord/playback.h)</requirement>
+ *   <requirement name="Library">k4arecord.lib</requirement>
+ *   <requirement name="DLL">k4arecord.dll</requirement>
+ * </requirements>
+ * \endxmlonly
+ */
+K4ARECORD_EXPORT uint64_t k4a_playback_data_block_get_timestamp_usec(k4a_playback_data_block_t data_block_handle);
```
wes-b: We now have device and and system timestamps. This should be namespaced with the timestamp being returned.

include/k4arecord/record.h
```diff
@@ -121,6 +127,156 @@ K4ARECORD_EXPORT k4a_result_t k4a_record_add_tag(k4a_record_t recording_handle,
...
+                                                                const char *track_name,
+                                                                const char *codec_id,
+                                                                const uint8_t *codec_context,
+                                                                size_t codec_context_size,
+                                                                const k4a_record_video_settings_t *track_settings);
+
+/** Adds custom subtitle tracks to the recording.
+ *
+ * Built-in subtitle tracks like the IMU track will be created automatically when the k4a_record_add_imu_track() API is
+ * called. This API can be used to add additional subtitle tracks to save custom data.
```
wes-b: This seems like \remarks info

include/k4arecord/record.h
```diff
@@ -202,6 +358,46 @@ K4ARECORD_EXPORT k4a_result_t k4a_record_write_capture(k4a_record_t recording_ha
...
+ *
+ * \param recording_handle
+ * The handle of a new recording, obtained by k4a_record_create().
+ *
+ * \param track_name
+ * The name of the custom track that the data is going to be written to.
+ *
+ * \param timestamp_usec
+ * The timestamp in microseconds for the custom track data.
+ *
```
wes-b: This timestamp should be namespaced

wes-b: Please apply this throughout where possible

include/k4arecord/types.h
```diff
@@ -48,6 +48,68 @@ K4A_DECLARE_HANDLE(k4a_record_t);
...
+ * \endxmlonly
+ */
+K4A_DECLARE_HANDLE(k4a_playback_data_block_t)
+
+/**
+ * @}
+ *
+ * \addtogroup Definitions
+ * @{
+ */
```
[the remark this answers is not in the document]
xthexder: This adds the below `#define`s to the docs here: https://microsoft.github.io/Azure-Kinect-Sensor-SDK/feature/support-custom-track-record-playback/group___definitions.html

include/k4arecord/record.h
xthexder: The recording files only have a single timestamp, which for captures ends up recording device timestamp.
Custom track data may not even come from a device, and the timestamp meaning is just user-defined. I'm not sure it makes sense to namespace these timestamps as device/system timestamps.

tests/RecordTests/UnitTest/playback_ut.cpp
xthexder: I'm not entirely sure what you mean by "validate inputs". Are you talking about testing all the `RETURN_VALUE_IF_ARG()` cases? Or something else.

wes-b: Tests that validate the inputs to your new functions. Like making sure passing in a NULL handle is a failure and not a crash. There should be tests that validate the full range of inputs each function can see.

include/k4arecord/record.h
wes-b: Playback API's basically mirror the SDK api's to get capture, images, and image info. The timestamp is one of them. Currently with the SDK the user can get two timestamps a device timestamp and a system timestamp. At some point you will need to add system timestamp and when that happens users will need to know the difference. So we should clarify what timestamp is being returned here and throughout. In 1.1 we deprecated k4a_image_get_timestamp_usec in favor of a new API k4a_image_get_device_timestamp_usec.

tests/RecordTests/UnitTest/playback_ut.cpp
xthexder: I think I've covered all the cases for valid inputs, as well as a couple of the less-obvious invalid input cases.
Testing things like NULL handle / input arguments for each function seems like a lot of extra code for not a lot of gain, considering every function already starts with something like `RETURN_VALUE_IF_HANDLE_INVALID(K4A_RESULT_FAILED, k4a_playback_t, playback_handle);`.
These cases are very unlikely to break in the future from code changes, and would have fairly low impact even if they don't immediately return.

I can go through and add tests for the full input space of each function, but that will have to wait until later next week, since I've got other things to work on in the mean time.

xthexder:
`k4a_playback_data_block_get_timestamp_usec()` has now been renamed to `k4a_playback_data_block_get_device_timestamp_usec()` to reflect the rest of the API changes.
I've updated the rest of the references to timestamp to say "device_timestamp", so fingers crossed 🤞 this is the final change to this PR.

PR #303, opened 2019-04-30
C# Build Integration

MsDerekMa:
This is integrating the C# build process into the Azure Pipeline build so we can get build and test running on the check-ins and nightly.


Brent-A:
I submitted a fix to feature/csharp that should resolve the build break. There may be more issues though since I am not using the same toolchain.

MsDerekMa:
Private fork was deleted.

PR #315, opened 2019-05-02
Linux binaries search local path for shared objects

### LINKED ISSUE #310: libk4a.so should search its local path for libdepthengine.so on Linux
On Windows, k4a.dll will look in its local path to load the depth engine plugin. On Linux this does not happen by default. This makes it painful to test newer versions of the depthengine plugin.

Set RPATH in the k4a shared object to be ORIGIN.

schultetwin1:
- Changes all binaries to also search for local path for shared objects. This emulates the same behavior on Windows. The shared object search path on Linux is now (list taken from: https://amir.rachum.com/blog/2016/09/17/shared-libraries/#runtime-search-path)

1. Directories in the LD_LIBRARY_PATH environment variable, which contains colon-separated list of directories (e.g., /path/to/libdir:/another/path)
2. Directory of the executable or shared object doing the loading
3. The list of directories in the file /etc/ld.so.conf. This file can include other files, but it is basically a list of directories - one per line.
4. Default system libraries - usually /lib and /usr/lib (skipped if compiled with -z nodefaultlib).

This better lines up with the Windows DLL search path (https://docs.microsoft.com/en-us/windows/desktop/dlls/dynamic-link-library-search-order#search-order-for-desktop-applications)

1. The directory from which the application loaded.
2. The system directory. Use the GetSystemDirectory function to get the path of this directory.
The 16-bit system directory. There is no function that obtains the path of this directory, but it is searched.
3. The Windows directory. Use the GetWindowsDirectory function to get the path of this directory.
4. The current directory.
5. The directories that are listed in the PATH environment variable. Note that this does not include the per-application path specified by the App Paths registry key. The App Paths key is not used when computing the DLL search path.

### I tested changes on:
- [ ] Windows
- [x] Linux

I ran functional tests on Linux to ensure the depthengine binary was picked up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,9 +71,19 @@ if ("${K4A_ENABLE_LEAK_DETECTION_CMAKE}" STREQUAL "1")
     add_definitions(-DK4A_ENABLE_LEAK_DETECTION)
 endif()
 
+# Sets the RUNPATH entry in the .dynamic section of an elf. RUNPATH is
+# interpreted by the linux loader as an additional path to search for shared
+# objects. $ORIGIN is a special setting telling the loader to search the path
+# relative to the exectuable.
 #
-set(CMAKE_BUILD_WITH_INSTALL_RPATH YES)
-set(CMAKE_INSTALL_RPATH "\$ORIGIN")
+# These specific settings tell the loader to search the directory of the
+# executable for shared objects. This is done on Linux to emulate the default
+# behavior of the Windows loader, which searches for DLLs in the path of the
+# executable.
+if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH YES)
+    set(CMAKE_INSTALL_RPATH "\$ORIGIN")
+endif()
 
 add_subdirectory(examples)
 add_subdirectory(src)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -71,6 +71,10 @@ if ("${K4A_ENABLE_LEAK_DETECTION_CMAKE}" STREQUAL "1")
     add_definitions(-DK4A_ENABLE_LEAK_DETECTION)
 endif()
 
+#
+set(CMAKE_BUILD_WITH_INSTALL_RPATH YES)
+set(CMAKE_INSTALL_RPATH "\$ORIGIN")
```
wes-b: >\$ORIGIN [](start = 25, length = 8)

Origin appear to be a token, so does that mean '\' is unnecessary? Example I see online are $ORIGIN\ which makes me think '\' is not needed.

CMakeLists.txt
```diff
@@ -71,6 +71,10 @@ if ("${K4A_ENABLE_LEAK_DETECTION_CMAKE}" STREQUAL "1")
     add_definitions(-DK4A_ENABLE_LEAK_DETECTION)
 endif()
 
+#
```
wes-b: Comment to to say RPATH of origin results in dlopen searching for binaries in the local directory first?

CMakeLists.txt
schultetwin1: The quotes are there because its a complicated string. Without the quotes, it was unclear to me if the "\\" is there for CMake reasons or not. The quotes help me understand, this full string is passed to the compiler. The backslash is needed to escape the "$" in the command line (see this mailing list for more info https://cmake.org/pipermail/cmake/2008-January/019290.html).

CMakeLists.txt
schultetwin1: Whoops, good call

PR #316, opened 2019-05-02
C# Build Integration

MsDerekMa:
- Standardize the project names to make it easier to find the binaries.
- Update the CMake process to support the Visual Studio generator.
- Addin the C# build and test to the Azure Pipeline.
- Moved test dependencies to be copied at build time instead of run time.
- Removed the Unit Test's runtime dependencies on K4A_BINARY_DIR and K4A_SOURCE_DIR.

### I tested changes on: 
- [x] Windows
- [ ] Linux


azure-pipelines.yml
```diff
@@ -233,6 +265,80 @@ jobs:
       testRunTitle: 'Unit Tests'
     condition: succeededOrFailed()
 
+  - powershell: |
```
Brent-A: We should at a comment indicating why we need to do this:
```
# Create a dummy version of the depth engine to satisfy MSBuild copy rules.
# This version is not needed for actual execution.
# In the future we should fetch the real version from a valid source.
```

azure-pipelines.yml
```diff
@@ -531,13 +638,24 @@ jobs:
       timeoutInMinutes: 15
       continueOnError: true
 
+    # TODO figure out the dependencies to be able to run.
```
Brent-A: If you file an issue, you should reference it here

src/csharp/K4a.sln
```diff
@@ -1,13 +1,13 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28809.33
+# Visual Studio 15
```
Brent-A: Is there a way that we can tell Visual Studio to not fight over this...

azure-pipelines.yml
```diff
@@ -233,6 +265,80 @@ jobs:
       testRunTitle: 'Unit Tests'
     condition: succeededOrFailed()
 
+  - powershell: |
+      Get-Content "$(Build.BinariesDirectory)\CMakeCache.txt" | where { $_ -match "CMAKE_GENERATOR:INTERNAL=(?<generator>.*)" }
+      if($Matches -ne $null)
+      {
+        if ($Matches["generator"] -eq "ninja")
```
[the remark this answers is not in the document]
MsDerekMa: I think so, that looks cleaner.

src/csharp/K4a.sln
MsDerekMa: This is a [bug with Visual Studio 2019](https://developercommunity.visualstudio.com/content/problem/516621/visual-studio-2019-is-still-changing-version-in-so.html) . There was a similar bug in [Visual Studio 2017] (https://developercommunity.visualstudio.com/content/problem/181684/visual-studio-writes-version-in-solution-file.html) that was fixed, so hopefully it gets fixed again. I believe this should list the lowest supported version, i.e. "Visual Studio 15".

PR #317, opened 2019-05-02
Remove Windows only sse

schultetwin1:
- Enable SSE for depth to xyz code

### I tested changes on:
- [ ] Windows
- [x] Linux

### DIFF
--- a/src/transformation/rgbz.c
+++ b/src/transformation/rgbz.c
@@ -791,6 +791,39 @@ k4a_buffer_result_t transformation_color_image_to_depth_camera_internal(
     return K4A_BUFFER_RESULT_SUCCEEDED;
 }
 
+// This is the same function as transformation_depth_to_xyz without the SSE
+// instructions. This code is kept here for readability.
+// static void transformation_depth_to_xyz(k4a_transformation_xy_tables_t *xy_tables,
+//                                         const void *depth_image_data,
+//                                         void *xyz_image_data)
+// {
+//     const uint16_t *depth_image_data_uint16 = (const uint16_t *)depth_image_data;
+//     int16_t *xyz_data_int16 = (int16_t *)xyz_image_data;
+//     int16_t x, y, z;
+
+//      for (int i = 0; i < xy_tables->width * xy_tables->height; i++)
+//     {
+//         float x_tab = xy_tables->x_table[i];
+
+//          if (!isnan(x_tab))
+//         {
+//             z = (int16_t)depth_image_data_uint16[i];
+//             x = (int16_t)(floorf(x_tab * (float)z + 0.5f));
+//             y = (int16_t)(floorf(xy_tables->y_table[i] * (float)z + 0.5f));
+//         }
+//         else
+//         {
+//             x = 0;
+//             y = 0;
+//             z = 0;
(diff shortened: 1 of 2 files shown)


src/transformation/rgbz.c
```diff
@@ -791,38 +791,6 @@ k4a_buffer_result_t transformation_color_image_to_depth_camera_internal(
...
-            x = 0;
-            y = 0;
-            z = 0;
-        }
-
-        xyz_data_int16[3 * i + 0] = x;
-        xyz_data_int16[3 * i + 1] = y;
-        xyz_data_int16[3 * i + 2] = z;
-    }
-}
```
mbleyer: I would rather not want to delete this code. It will be useful for people to understand what the SSE version is doing.

schultetwin1:
@mbleyer this build break brings up a good error. On line 799 of rgbz.c we do the following cast:

```c
__m128 *x_table_m128 = (__m128 *)xy_tables->x_table;
```

`xy_tables->x_table` is a pointer to a float, which means it's pointing to a 4-byte aligned data member. We are casting that to a pointer to a `__m128`, meaning a pointer to a 16-byte aligned data member.

I've added assumptions in to get around this but is this something we should worry about?

mbleyer:
> @mbleyer this build break brings up a good error. On line 799 of rgbz.c we do the following cast:
> __m128 *x_table_m128 = (__m128 *)xy_tables->x_table;
> xy_tables->x_table is a pointer to a float, which means it's pointing to a 4-byte aligned data member. We are casting that to a pointer to a __m128, meaning a pointer to a 16-byte aligned data member.
> I've added assumptions in to get around this but is this something we should worry about?

You are right. Maybe we should open a P3 bug on this.

src/transformation/rgbz.c
schultetwin1: Sounds good, put it back in

schultetwin1:
> > @mbleyer this build break brings up a good error. On line 799 of rgbz.c we do the following cast:
> > __m128 *x_table_m128 = (__m128 *)xy_tables->x_table;
> > xy_tables->x_table is a pointer to a float, which means it's pointing to a 4-byte aligned data member. We are casting that to a pointer to a __m128, meaning a pointer to a 16-byte aligned data member.
> > I've added assumptions in to get around this but is this something we should worry about?
> 
> You are right. Maybe we should open a P3 bug on this.

Done, I've opened issue #318 and assigned it to you.

Brent-A:
Does this change just ensure that the compiler is set to support SSE4.1? Don't we need to perform a runtime check as well to determine if the platform we are executing on supports those instructions, or we get a runtime failure?

schultetwin1:
> Does this change just ensure that the compiler is set to support SSE4.1? Don't we need to perform a runtime check as well to determine if the platform we are executing on supports those instructions, or we get a runtime failure?

Good point, we should have a better failure than just a CPU fault on an unsupported instruction. Since this is already a problem (our depth engine already uses SSE4.1 instructions), I have created an issue to add that in a different PR. #324

schultetwin1:
@mbleyer or @cdedmonds can you guys take a look at this today?

Issue #240, opened 2019-04-15
Add documentation on acceptable RGB camera manual exposure values


mffigueroa:
When looking at either the [Kinect hardware specs](https://review.docs.microsoft.com/en-us/azurekinect/azure-kinect-devkit?branch=master) or the [k4a_color_control_command_t documentation](https://review.docs.microsoft.com/en-us/azurekinect/api/k4a-color-control-command-t?branch=master), it is not clear what the acceptable values are for the RGB camera exposure time setting. As a user, the experience then involves trying different values randomly just to see what works.

DrMaxP:
Any ideas on when the hardware specs and this documentation will be accessible to external users?

joylital:
We have some specs here: aka.ms/kinectdocs. Take a look and let us know what additional information is needed to help us improve that.

DrMaxP:
Ok. So general hardware capabilities I haven't seen addressed are:

Brent-A:
Thanks for your interest @DrMaxP! We started a new FAQ list which I hope answers your questions in issue #338.

wes-b:
#362 is related, waiting to hear back from the factory on HW capabilities.

tesych:
the table is not added to [hardware spec](https://docs.microsoft.com/azure/Kinect-dk/hardware-specification#rgb-camera-exposure-time-values)

PR #319, opened 2019-05-02
New API to query color camera control capabilities.

### LINKED ISSUE #216: New API for querying default color control values and ranges

MinshikPark:
- Add k4a_device_get_color_control_capabilities() API to query color camera control capabilities

### I tested changes on: 
- [x] Windows
- [x] Linux

color functional test (color_ft) include testing this feature.

### DIFF
--- a/include/k4a/k4a.h
+++ b/include/k4a/k4a.h
@@ -1211,7 +1211,7 @@ K4A_EXPORT k4a_result_t k4a_device_get_version(k4a_device_t device_handle, k4a_h
  * \param command
  * Color sensor control command.
  *
- * \param support_auto
+ * \param supports_auto
  * Location to store whether the color sensor's control support auto mode or not.
  * true if it supports auto mode, otherwise false.
  *
@@ -1245,7 +1245,7 @@ K4A_EXPORT k4a_result_t k4a_device_get_version(k4a_device_t device_handle, k4a_h
  */
 K4A_EXPORT k4a_result_t k4a_device_get_color_control_capabilities(k4a_device_t device_handle,
                                                                   k4a_color_control_command_t command,
-                                                                  bool *support_auto,
+                                                                  bool *supports_auto,
                                                                   int32_t *min_value,
                                                                   int32_t *max_value,
                                                                   int32_t *step_value,
--- a/include/k4a/k4a.h
+++ b/include/k4a/k4a.h
@@ -1203,7 +1203,57 @@ K4A_EXPORT k4a_buffer_result_t k4a_device_get_serialnum(k4a_device_t device_hand
  */
 K4A_EXPORT k4a_result_t k4a_device_get_version(k4a_device_t device_handle, k4a_hardware_version_t *version);
 
-/** Get the Azure Kinect color sensor control value.
+/** Get the Azure Kinect color sensor control capabilities.
+ *
+ * \param device_handle
+ * Handle obtained by k4a_device_open().
+ *
+ * \param command
+ * Color sensor control command.
+ *
+ * \param support_auto
+ * Location to store whether the color sensor's control support auto mode or not.
+ * true if it supports auto mode, otherwise false.
+ *
+ * \param min_value
+ * Location to store the color sensor's control minimum value of /p command.
+ *
+ * \param max_value
+ * Location to store the color sensor's control maximum value of /p command.
+ *
+ * \param step_value
+ * Location to store the color sensor's control step value of /p command.
+ *
+ * \param default_value
+ * Location to store the color sensor's control default value of /p command.
+ *
+ * \param default_mode
+ * Location to store the color sensor's control default mode of /p command.
+ *
+ * \returns
+ * ::K4A_RESULT_SUCCEEDED if the value was successfully returned, ::K4A_RESULT_FAILED if an error occurred
+ *
+ * \relates k4a_device_t
+ *
+ * \xmlonly
+ * <requirements>
+ *   <requirement name="Header">k4a.h (include k4a/k4a.h)</requirement>
+ *   <requirement name="Library">k4a.lib</requirement>
+ *   <requirement name="DLL">k4a.dll</requirement>
+ * </requirements>
+ * \endxmlonly
+ */
+K4A_EXPORT k4a_result_t k4a_device_get_color_control_capabilities(k4a_device_t device_handle,
+                                                                  k4a_color_control_command_t command,
+                                                                  bool *support_auto,
+                                                                  int32_t *min_value,
+                                                                  int32_t *max_value,
+                                                                  int32_t *step_value,
+                                                                  int32_t *default_value,
+                                                                  k4a_color_control_mode_t *default_mode);
+
+/** Get the Azure Kinect color sensor control
+ * value.
  *
  * \param device_handle
  * Handle obtained by k4a_device_open().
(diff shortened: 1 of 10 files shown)


tests/ColorTests/FunctionalTest/color_ft.cpp
```diff
@@ -642,8 +650,15 @@ TEST_P(color_control_test, control_test)
 
     if (as.command == K4A_COLOR_CONTROL_EXPOSURE_TIME_ABSOLUTE)
     {
+#ifdef _WIN32
+        // Windows only support exposure time value as log base 2 seconds.
```
wes-b: We should talk about this, what will it take to remove #ifdef's in the code? This means to have  different behavior for different platforms, and I would prefer to avoid that. This also means we have to document the behavior too.

include/k4a/k4a.h
```diff
@@ -1203,7 +1203,57 @@ K4A_EXPORT k4a_buffer_result_t k4a_device_get_serialnum(k4a_device_t device_hand
...
+ * <requirements>
+ *   <requirement name="Header">k4a.h (include k4a/k4a.h)</requirement>
+ *   <requirement name="Library">k4a.lib</requirement>
+ *   <requirement name="DLL">k4a.dll</requirement>
+ * </requirements>
+ * \endxmlonly
+ */
+K4A_EXPORT k4a_result_t k4a_device_get_color_control_capabilities(k4a_device_t device_handle,
+                                                                  k4a_color_control_command_t command,
+                                                                  bool *support_auto,
```
wes-b: >support [](start = 72, length = 7)

support[s]_auto

include/k4a/k4a.h
```diff
@@ -1203,7 +1203,57 @@ K4A_EXPORT k4a_buffer_result_t k4a_device_get_serialnum(k4a_device_t device_hand
...
+                                                                  int32_t *min_value,
+                                                                  int32_t *max_value,
+                                                                  int32_t *step_value,
+                                                                  int32_t *default_value,
```
wes-b: Can we have a functional test that reads these and confirms they are always the same for linux and windows?

We expect this to be consistent through out.

tests/ColorTests/FunctionalTest/color_ft.cpp
MinshikPark: I will think how we can eliminate #ifdef for this.
And documentation bug is opened #240

include/k4a/k4a.h
MinshikPark: I will rename them.

include/k4a/k4a.h
MinshikPark: Do you mean the values returned this function from Linux are same with from Windows?
Or, in certain platform, consecutive call of this function returns same values with previous call?

wes-b: after a device reset, I would expect the values returned by the device would be exactly the same. And as a result this API should be consistent between the two. Differences between UVC and MF should be fixed in code so they remain consistent.

---
In reply to: [280931778](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/pull/319#discussion_r280931778) [](ancestors = 280931778)

tests/ColorTests/FunctionalTest/color_ft.cpp
wes-b: The goal is that this enhancement doesn't result in more complicated documentation.

---
In reply to: [280930242](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/pull/319#discussion_r280930242) [](ancestors = 280930242)

MinshikPark: I think this #ifdef does not make more complicated documentation, but it support #240 documentation for better understanding.

include/k4a/k4a.h
MinshikPark: That's good idea.
Just curious that because there is no guarantee that the functional test will be run after device reset, is it a good idea to reset the device in the functional test?
Do we have an API to reset the device?

Thanks.

MinshikPark: Added default value checking in the functional test with failure message if device was not reset after control value has changed.

src/color/mfcamerareader.cpp
```diff
@@ -539,14 +539,165 @@ void CMFCameraReader::Stop()
     }
 }
 
+k4a_result_t CMFCameraReader::GetCameraControlCapabilities(const k4a_color_control_command_t command,
+                                                           color_control_cap_t &capabilities)
```
williampMSFT: >& [](start = 79, length = 1)

avoid non-const lvalue references if possible - it makes it easier to tell that the thing you're passing in is being manipulated

src/color/mfcamerareader.cpp
```diff
@@ -539,14 +539,165 @@ void CMFCameraReader::Stop()
...
+    }
+    break;
+    case K4A_COLOR_CONTROL_AUTO_EXPOSURE_PRIORITY:
+    {
+        supportAuto = false;
+        minValue = 0;
+        maxValue = 0;
+        stepValue = 0;
+        defaultValue = 0;
+        LOG_WARNING("K4A_COLOR_CONTROL_AUTO_EXPOSURE_PRIORITY is deprecated and do nothing.");
```
williampMSFT: >do [](start = 80, length = 2)

nit: "does"

src/color/mfcamerareader.cpp
```diff
@@ -539,14 +539,165 @@ void CMFCameraReader::Stop()
...
+                                 &supportAuto,
+                                 &minValue,
+                                 &maxValue,
+                                 &stepValue,
+                                 &defaultValue);
+
+        // Convert KSProperty exposure time value to micro-second unit
+        minValue = (LONG)(exp2f((float)minValue) * 1000000.0f);
+        maxValue = (LONG)(exp2f((float)maxValue) * 1000000.0f);
+        stepValue = 1;
```
williampMSFT: >stepValue = 1; [](start = 8, length = 14)

is this true?  I thought stepping on this was nonlinear/exponential?  I don't know if we can even express that with this API, so I'm not entirely sure what to do with this one.  Probably at minimum we want some sort of comment if this isn't actually true

MinshikPark: I agree, this stepValue is actually not true for Media Foundation, but its "kind of true" in K4A API surface because it accept even its truncated into exponential value.
The other option for this API might be returning list of available values, but it's hard to make generic API for other controls and Linux version of API.
Any other better idea will be very appreciated.

src/color/mfcamerareader.cpp
MinshikPark: Thanks, I will fix them.

src/color/mfcamerareader.cpp
MinshikPark: I will use pointer.

src/color/mfcamerareader.cpp
williampMSFT: I see two possible approaches:

A)
I think setting it to 1 is fine since it'll still succeed, but we might want to stick a comment in here explaining that if you step by 1 it'll actually go up by a power of 2 so if someone gets confused and starts debugging, they'll understand why we're hard-setting it to 1 here.

B)
The only way I see to do a generic API here is to have an extra out parameter that's an enum of step types, something like 
```
enum K4A_COLOR_CONTROL_STEP_TYPE
{
    K4A_COLOR_CONTROL_STEP_TYPE_LINEAR,
    K4A_COLOR_CONTROL_STEP_TYPE_EXPONENTIAL
};
```

and then set everything except this one to linear.  This has the advantage of being more accurate, but adds complexity to every other color control that might not be worth it.

I don't have strong feelings on which approach to take here, but if we go with A) I do think adding a comment there explaining how it actually snaps to 1000000/[a power of 2] will probably save someone a fair amount of debugging at some point

---
In reply to: [281310570](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/pull/319#discussion_r281310570) [](ancestors = 281310570)

MinshikPark: I added some comment on it. Thanks.

PR #327, opened 2019-05-04
Enabling C# functional tests during build

### LINKED ISSUE #322: Enable C# functional tests in the build
This is related to getting C# support enabled, #136. The functional tests are missing a dependency when attempting to run and so they fail. This issue is to track getting the dependencies resolved and the functional tests running.

MsDerekMa:
- Updating the Azure Pipeline YML to correctly install VSTest and run the C# functional tests.
- Adding C# binaries to the artifacts so that the functional tests can use them.
- Adding extra path delimiters to the projects so that they are resilient to not having a trailing slash in the Base Output Path.

### I tested changes on: 
- [ ] Windows
- [ ] Linux


azure-pipelines.yml
```diff
@@ -296,35 +296,23 @@ jobs:
     displayName: 'Build C# Wrappers'
     inputs:
       solution: '$(Build.SourcesDirectory)\src\csharp\K4a.sln'
-      msbuildArgs: '/p:BaseOutputPath="$(Build.BinariesDirectory)\$(MSBuildConfiguration)\\"'
+      msbuildArgs: '/p:BaseOutputPath="$(Build.BinariesDirectory)\bin\$(MSBuildConfiguration)\\"'
```
schultetwin1: Why do we need the \\\\ after $(MSBuildConfiguration)?

MsDerekMa: It was so that we don't escape the quote. But I did fix the csproj files so that we don't need the trailing slash. I could remove it.

PR #350, opened 2019-05-15
Added k4a_image_get/set_system_timestamp_nsec

### LINKED ISSUE #198: Provide method to convert device timestamps to system timestamps
**Tasks**
* [x] 1. API Design covering Windows and Linux, as well as Live and Playback
* [x] 2. Implement simple time sync by correlating PTS with CPU time at frame receipt
* [ ] 3. Implement precision time sync by reading PTS/USB frame correlation from firmware

**API design considerations**

* Hardware PTS timer has a resolution of 90kHz and resets to zero at boot and may reset during camera on/off transitions.
* Hardware PTS may drift in respect to the host clock (we should double check this assertion, and determine the worst case drift rate for long running devices).
* Because the PTS timer resets between camera on/off cycles, a function like `k4a_device_get_systemtime_from_timestamp()` may not be appropriate, or may have to be scoped to only being valid for timestamps received since the previous call to `k4a_device_start_cameras()`, since the same device timestamp may map to multiple absolute times.
* Do we need to convert to an absolute time (with date) or accurate boot specific timer?
* Will time conversion be available for recordings? If it is boot specific like `QueryPerformanceCounter` does it really make sense? Recordings taken on Linux should presumably be able to be read on Windows and visa versa with appropriate time APIs.

**Units common to platforms**

*Windows*

* `QueryPerformanceCounter`: highest precision time available, resets on each boot of PC. 
* `FILETIME`: 100-nanosecond intervals since 1601-01-01
* `SYSTEMTIME`: 1 millisecond intervals (modern systems should report in 1ms resolution)

*Linux*

* `clock_gettime(CLOCK_MONOTONIC, …)`: Highest precision available. Resets on each boot of the PC. Subject to NTP adjustment and doesn't tick during suspsend
* `clock_gettime(CLOCK_MONOTONIC_RAW, …)`: Highest precision available. Resets on each boot of the PC. Not adjusted by NTP. Only available on kernel 2.6.28 and newer.
* `clock_gettime(CLOCK_REALTIME, …)`: 1 nanosecond intervals of PC dependent resolution as an absolute time (subject to clock adjustments)
* `time_t`: seconds since 1970-01-01

**Other discussion on common timestamping**

It looks like ros2/rcutils#43 and ziglang/zig#933 both concluded that CLOCK_MONOTINIC and QueryPerformanceCounter were appropriate units for their OS abstraction. If this forms the community consensus we may want to ensure that our APIs are aligned.

wes-b:
- Added k4a_image_get_system_timestamp_nsec() & k4a_image_set_system_timestamp_nsec().
- Added k4a_image_get_device_timestamp_usec().
- Added k4a_image_set_exposure_usec().
- Deprecated k4a_image_get_timestamp_usec() & k4a_image_set_timestamp_usec().
- Deprecated k4a_image_set_exposure_time_usec().
- System Timestamp on Windows is based on QueryPerformanceCounter, while Linux is based on clock_gettime(CLOCK_MONOTONIC). There is a nice writeup here: [ros2/rcutils#43](https://github.com/ros2/rcutils/issues/43).
- libuvc needed to be updated to capture the time at the completion of last USB packet. So libuvc was converted from a patch to a fork. 

### I tested changes on: 
- [x] Windows
- [x] Linux


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Change Log
 
+### v1.1.0
```
xthexder: There's a large amount of other changes that are missing from the 1.1.0 changelog. Should we maybe move this part to a separate PR?

.gitmodules
```diff
@@ -35,3 +35,7 @@
 [submodule "extern/libusb/src"]
 	path = extern/libusb/src
 	url = https://github.com/libusb/libusb
+[submodule "extern/libuvc/src"]
+	path = extern/libuvc/src
+	url = https://github.com/wes-b/libuvc.git
```
williampMSFT: >wes-b [](start = 26, length = 5)

We might want to have the libuvc fork owned by the Microsoft corporate account to increase our bus factor

CHANGELOG.md
wes-b: Actually this change didn't make 1.1, so I will rev it to 1.2.0 and add a bit about 1.1.0. Most of 1.1.0 was bug fixes.

include/k4a/k4a.hpp
```diff
@@ -339,11 +348,11 @@ class image
 
     /** Set the image's timestamp in microseconds
      *
-     * \sa k4a_image_set_timestamp_usec
+     * \sa k4a_image_set_device_timestamp_usec
      */
     void set_timestamp(std::chrono::microseconds timestamp) noexcept
```
williampMSFT: do we want a set_device_timestamp and set_system_timestamp here now?

include/k4a/k4a.hpp
```diff
@@ -301,13 +301,22 @@ class image
         return k4a_image_get_stride_bytes(m_handle);
     }
 
-    /** Get the image timestamp in microseconds
+    /** Get the image's device timestamp in microseconds
      *
-     * \sa k4a_image_get_timestamp_usec
+     * \sa k4a_image_get_device_timestamp_usec
      */
-    std::chrono::microseconds get_timestamp() const noexcept
```
williampMSFT: >std::chrono::microseconds get_timestamp() const noexcept [](start = 4, length = 56)

If this change makes it in for 1.1, we don't need to preserve this (this wrapper wasn't in 1.0), but if we're punting to 1.2, deleting this is a breaking API change - we may need to keep it for backwards compat and mark it deprecated :(

.gitmodules
wes-b: I agree with this long term. In the short term I have pushed a PR back to the owner to see if I get any traction with them taking a fix. There is not too much action over at https://github.com/ktossell/libuvc, so if that doesn't work out we will work to get this forked under Microsoft. For now this change is unblocked while we explore our libuvc options.

williampMSFT: Ah, okay, I didn't realize the fork was a temporary thing pending a PR being accepted.  In that case, this is fine.

---
In reply to: [284937090](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/350#discussion_r284937090) [](ancestors = 284937090)

include/k4a/k4a.hpp
wes-b: yes, for the playback case

include/k4a/k4a.hpp
wes-b: Good call, I am going to see if we can fix this API in our 1.1 release 

---
In reply to: [284936947](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/pull/350#discussion_r284936947) [](ancestors = 284936947)

williampMSFT:
>                                                        k4a_imu_sample_t *imu_sample,

This may be out of scope for this change, but figured I'd mention it:
k4a_imu_sample_t has (presumably device) timestamps embedded in it - do we need to provide a facility for getting system timestamps for IMU samples?

---
Refers to: include/k4a/k4a.h:255 in ac64c8d. [](commit_id = ac64c8da77041206a614015fa7a524be0fe2e89c, deletion_comment = False)

include/k4a/k4a.h
```diff
@@ -886,13 +947,76 @@ K4A_EXPORT uint32_t k4a_image_get_white_balance(k4a_image_t image_handle);
...
+ * \remarks
+ * Set a timestamp of 0 to indicate that the timestamp is not valid.
+ *
+ * \remarks
+ * The device timestamp represents the mid-point of exposure of the image, as captured by the hardware.
+ *
+ * \relates k4a_image_t
+ *
+ * \deprecated
+ * Deprecated starting in 1.1.0. Please use k4a_image_set_device_timestamp_usec().
```
schultetwin1: Just verifying this should still say deprecated in 1.1.0?

include/k4a/k4a.h
```diff
@@ -886,13 +947,76 @@ K4A_EXPORT uint32_t k4a_image_get_white_balance(k4a_image_t image_handle);
...
+ *
+ * \param timestamp_usec
+ * Device timestamp of the image in microseconds.
+ *
+ * \remarks
+ * Use this function in conjunction with k4a_image_create() or k4a_image_create_from_buffer() to construct a
+ * \ref k4a_image_t.
+ *
+ * \remarks
+ * Set a timestamp of 0 to indicate that the timestamp is not valid.
```
schultetwin1: In `k4a_image_get_device_timestamp_usec` we mention that a timestamp of 0 is valid. Here we mention that its not valid. Is there a difference between a created frame and a recorded frame?

include/k4a/k4a.h
```diff
@@ -901,6 +1025,35 @@ K4A_EXPORT uint32_t k4a_image_get_iso_speed(k4a_image_t image_handle);
  * \remarks
  * Set a timestamp of 0 to indicate that the timestamp is not valid.
  *
+ * \remarks
+ * The system timestamp is a high performance and increasing clock (from boot). The timestamp represents the time
+ * immediately after the image buffer was ready by the host PC.
```
schultetwin1:
```suggestion
 * immediately after the image buffer was read by the host PC.
```

include/k4ainternal/common.h
```diff
@@ -24,7 +24,8 @@ typedef struct _guid_t
 #define MAX_FPS_IN_MS (33) // 30 FPS
 
 #define COUNTOF(x) (sizeof(x) / sizeof(x[0]))
-
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
```
schultetwin1: To avoid double evaluation bugs in the future:

```suggestion
#define MIN(a, b) \
({ __typeof__ (a) _a = (a); \
   __typeof__ (b) _b = (b); \
   (_a < _b) ? _a : _b; })
```

include/k4ainternal/common.h
```diff
@@ -24,7 +24,8 @@ typedef struct _guid_t
...
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
```
schultetwin1: To avoid double evaluation bugs in the future:

```suggestion
#define MAX(a, b) \
({ __typeof__ (a) _a = (a); \
   __typeof__ (b) _b = (b); \
   (_a > _b) ? _a : _b; })
```

src/color/mfcamerareader.cpp
```diff
@@ -1056,10 +1056,22 @@ STDMETHODIMP CMFCameraReader::OnReadSample(HRESULT hrStatus,
...
                 {
-                    image_set_timestamp_usec(image, K4A_90K_HZ_TICK_TO_USEC(pFrameContext->GetPTSTime()));
+                    // Read the QPC value MF attached to the sample when it was received.
+                    unsigned long long mf_device_time_100nsec = 0;
+                    if (FAILED(hr = pSample->GetUINT64(MFSampleExtension_DeviceTimestamp, &mf_device_time_100nsec)))
+                    {
+                        result = K4A_RESULT_FAILED;
+                        LOG_ERROR("IMFSample::GetUINT64(MFSampleExtension_DeviceTimestamp) failed; hr=0x%08X", hr);
+                    }
+                    image_set_system_timestamp_nsec(image, mf_device_time_100nsec * 100);
```
schultetwin1: Should we only set the system_timestamp if the result `pSample->GetUINT64` succeeded?

src/image/image.c
```diff
@@ -293,18 +305,59 @@ uint32_t image_get_iso_speed(k4a_image_t image_handle)
...
+        // Rollover happens, for a 1MHz Freq, when qpc.QuadPart > 0x003F FFFF FFFF FFFF; ~571 Years after boot.
+        image->sys_timestamp_nsec = qpc.QuadPart / freq.QuadPart * 1000000000;
+        image->sys_timestamp_nsec += qpc.QuadPart % freq.QuadPart * 1000000000 / freq.QuadPart;
+    }
+#else
+    struct timespec ts_time;
+    result = K4A_RESULT_FROM_BOOL(clock_gettime(CLOCK_MONOTONIC, &ts_time) == 0);
+    if (K4A_SUCCEEDED(result))
+    {
+        // Rollover happens about ~136 years after boot.
```
schultetwin1: I love this comment

tests/ColorTests/FunctionalTest/color_ft.cpp
```diff
@@ -161,6 +165,28 @@ TEST_P(color_functional_test, color_streaming_test)
             break;
         }
 
+        if (ts_init)
+        {
+            // Ensure the device and system time stamps are increasing, image might get dropped, which is ok for this
+            // portion of the test.
+            uint64_t ts;
+            ts = k4a_image_get_device_timestamp_usec(image);
+            ASSERT_GT(ts, ts_d); // We don't test upper max, because we expect the next sample to be larger, this also
```
schultetwin1: Since we don't initially set `ts_d` to a value I'd be very impressed if all of our compilers pass by this without complaining about a possible use before assignment.

tests/TestUtil/Capture.cpp
```diff
@@ -118,7 +117,7 @@ static void image_stream_callback(k4a_result_t result, k4a_image_t image_handle,
     }
     else
     {
-        printf("%16d                    ", (int)image_get_timestamp_usec(image_handle));
+        printf("%16d                    ", (int)image_get_device_timestamp_usec(image_handle));
```
schultetwin1: Maybe we should change this to not cast to an int and print out the uint64_t?
```suggestion
        printf("%16" PRId64 "                    ", image_get_device_timestamp_usec(image_handle));
```

tests/TestUtil/Capture.cpp
```diff
@@ -136,7 +135,7 @@ static void image_stream_callback(k4a_result_t result, k4a_image_t image_handle,
         }
         else
         {
-            fprintf(p_file, "%d, ", (int)image_get_timestamp_usec(image_handle));
+            fprintf(p_file, "%d, ", (int)image_get_device_timestamp_usec(image_handle));
```
schultetwin1: Same comment here about the cast to (int)

```suggestion
            fprintf(p_file, PRId64 ", ", image_get_device_timestamp_usec(image_handle));
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 ## Change Log
 
+### v1.2.0
+
+* Added new API's k4a_image_get_device_timestamp_usec(), k4a_image_get_system_timestamp_usec(), 
+k4a_image_set_device_timestamp_usec(), k4a_image_set_system_timestamp_usec(), and k4a_image_set_exposure_usec().
```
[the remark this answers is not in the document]
wes-b: This is just an artifact of the line wrapping while looking at the raw MD file. It is rendered as you would expect it to.

src/image/image.c
wes-b: :)

include/k4a/k4a.h
```diff
@@ -789,20 +789,81 @@ K4A_EXPORT int k4a_image_get_height_pixels(k4a_image_t image_handle);
...
+ * Returns the device timestamp of the image. Timestamps are recorded by the device and represent the mid-point of
+ * exposure. They may be used for relative comparison, but their absolute value has no defined meaning.
  *
  * \returns
  * If the \p image_handle is invalid or if no timestamp was set for the image,
  * this function will return 0. It is also possible for 0 to be a valid timestamp originating from the beginning
  * of a recording or the start of streaming.
  *
+ * \deprecated
+ * Deprecated starting in 1.1.0. Please use k4a_image_get_device_timestamp_usec().
```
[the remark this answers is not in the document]
wes-b: Good Catch

include/k4a/k4a.h
wes-b: Nope, good catch

include/k4ainternal/common.h
schultetwin1: We should do the same for MIN (currently this change is only on MAX)

wes-b:
__typeof__ is not C, I got lucky building before because that macro was never used in C code. I am reverting the change.

schultetwin1:
> `__typeof__` is not C, I got lucky building before because that macro was never used in C code. I am reverting the change.

Ugh, its a GNU extension. Good find. If we are never using these MIN and MAX macros in C code, then should we include them? C++ provides a much safer std::max and std::min that we can use.

schultetwin1:
> > `__typeof__` is not C, I got lucky building before because that macro was never used in C code. I am reverting the change.
> 
> Ugh, its a GNU extension. Good find. If we are never using these MIN and MAX macros in C code, then should we include them? C++ provides a much safer std::max and std::min that we can use.

Talked with Wes offline, we are using these macros in C code. Looks good to me!

PR #365, opened 2019-05-22
Add artifacts to release document

### LINKED ISSUE #361: Document released artifacts
We need a clear definition of all of the artifacts that are released in various packages.

In [releasing.md](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/blob/develop/docs/releasing.md) we should include a table with the following columns:
* File (including a path, e.g. "windows-desktop/lib/release/amd64/k4a.dll")
* Source (e.g. "Azure-Kinect-Sensor-SDK build")
* Packages (e.g. "msi, nuget, deb")

MsDerekMa:
- Creating a table that contains all of our build artifacts to help get an understanding of what is being shipped where.
- Calling out future .NET binaries.

### I tested changes on: 
- [x] Windows
- [ ] Linux


docs/releasing.md
```diff
@@ -79,8 +79,68 @@ The pipeline will:
...
 this [GitHub repo](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK).
 
-[Release tags](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/releases) are used by Microsoft to label commits 
+[Release tags](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/releases) are used by Microsoft to label commits
 used to create a released build. Checkout the commit that matches the release label to build the desired release version.
+
+## Artifacts and Packaging
+The following table contains all artifacts that are released in installers and packages. The file paths listed in the table are only representative, within any given package they will be different to match the standards and conventions for those packages. Each unique released artifact is listed in this table only once, although they will potentially appear duplicate times in packages or installers.
+
+File                                                           | MSI                | NuGet              | libk4a             | libk4a-dev         | libk4a-tools
```
schultetwin1: What about changing the debian packages to be what they are (like "MSI" and "Nuget") instead of the package name?

```suggestion
File                                                           | MSI                | NuGet              | Runtime Debian Package             | Development Debian Package         | Tooling Debian Package
```

docs/releasing.md
```diff
@@ -79,8 +79,68 @@ The pipeline will:
...
-The Azure Kinect SDK installer provides both open and closed source binaries. The open source binaries are built from 
+The Azure Kinect SDK installer provides both open and closed source binaries. The open source binaries are built from
 this [GitHub repo](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK).
 
-[Release tags](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/releases) are used by Microsoft to label commits 
+[Release tags](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/releases) are used by Microsoft to label commits
 used to create a released build. Checkout the commit that matches the release label to build the desired release version.
+
+## Artifacts and Packaging
+The following table contains all artifacts that are released in installers and packages. The file paths listed in the table are only representative, within any given package they will be different to match the standards and conventions for those packages. Each unique released artifact is listed in this table only once, although they will potentially appear duplicate times in packages or installers.
```
schultetwin1: I know we talked about this in person, but no conclusion was reached. What about just listing file names and not file paths? I understand we might have some overlap between release and debug binaries, but I think it would greatly simplify both the readability of the table and the size of the table. If the user wants to understand if the package contains both release and debug. Or both 32-bit and 64-bit they can dump the contents of the package.

Brent-A: Derek had tried listing the binaries without paths and instead just qualifying them a comment indicating what type they are if there is potential ambiguity and it just gets hard to read. Since the file names alone are not distinct, I think it's important to have a place where all the relevant qualifies are listed.

For example, I want to be real explicit about showing that there is only ever one version of "k4aTargets.cmake", but there are two versions "Microsoft.Azure.Kinect.Sensor.targets" depending on where you look.

We could replace the `/` with `#` or some other separator to explicitly show that these are categories and not paths, but I think it just makes things less readable. I would also propose that in the MSI we literally use these exact paths to keep things simple.

schultetwin1: Brent and I talked in person. Since the `/` shows a more hierarchic view of the world I'm fine sticking with that for now.

PR #368, opened 2019-05-23
Revert "Install k4arecorder k4aviewer"

wes-b:
Reverts microsoft/Azure-Kinect-Sensor-SDK#367


wes-b:
The PR schultewin1 created that I accidentally submitted was ready to be submitted and does not need to be reverted.

PR #402, opened 2019-06-10
Fixed some spelling and grammar and removed hard line breaks

codefoster:
## Fixes #
Just read that PR's should fix a Triage Approved issue. Submitting anyway. Feel free to delete it.

Fixed some spelling and grammar issues in the `usage.md` file. Also removed extrinsic hard line breaks.

Did not test. Only minor changes to a markdown file.


codefoster:
I made sure the line breaks change was in a separate commit, but forgot that it would obscure the other changes, because of the GH change render issue you mentioned.
Regarding line break policy, do you suppose my guess would be accurate that 98.75% of people would be using an editor that supports line breaks? If that's the case, then I'm sure there's more pain in keeping them. Just my $0.02.
I can add the line breaks back in early next week.
Thanks.

PR #425, opened 2019-06-20
Float2 and Float3 replaced with Vector2 and Vector3

gllebede:
Custom vector structures (Float2 and Float3) replaced with a .net implementation from System.Numerics namespace.

### I tested changes on: 
- [x] Windows
- [ ] Linux

Tested with existing unit tests and via manually running provided sample applications.


src/csharp/Microsoft.AzureKinect.UnitTests/StructureTests.cs
```diff
@@ -8,9 +9,9 @@ namespace Microsoft.AzureKinect.UnitTests
     public class StructureTests
     {
         [Test]
-        public void Float2()
+        public void Vector2()
```
xthexder: This can probably all be removed. I don't think we need to test system libraries in our code.

gllebede: Will remove it then. Initially I've decided to keep them in place to check compatibility.

PR #426, opened 2019-06-20
Make examples test cases

spschul:
- Makes the examples test cases.

### I tested changes on: 
- [x] Windows
- [x] Linux


examples/transformation/main.cpp
```diff
@@ -195,14 +195,22 @@ static int capture(std::string output_dir, uint8_t deviceId = K4A_DEVICE_DEFAULT
     }
 
     // Compute color point cloud by warping color image into depth camera geometry
+#ifdef _WIN32
```
Brent-A: Windows will actually accept forward slashes as valid path separators. We may be able to drop the #if logic here.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,229 @@
...
+                                          << *regex_iter << "\", but never saw it.";
+}
+
+class examples_ft : public ::testing::Test
+{
+protected:
+    void SetUp() override
+    {
+#ifdef _WIN32
+        // _putenv_s("K4A_ENABLE_LOG_TO_STDOUT", "0");
```
Brent-A: Why is this commented out? If there's a reason this should behave differently on different platforms, a comment would be good here.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+    // TODO fflush needed here?
+    std::string formatted_command = executable_path;
+    formatted_command += " " + args;
+    if (!output_path.empty())
+    {
+        formatted_command += " 2>&1 > " + output_path;
+    }
+    std::cout << formatted_command << std::endl;
+#ifdef _WIN32
```
Brent-A: This function could probably just have a single #if block

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <vector>
+#include <k4a/k4a.h>
+#include <utcommon.h>
+#include <gtest/gtest.h>
+
+const static std::string TEST_TEMP_DIR = "examples_test_temp";
+
+// run the specified executable and record the output to output_path
+// if output_path is empty, just output to stdout
+static void run_and_record_executable(const std::string &executable_path,
```
Brent-A: Since this function is just appending the executable_path and args, perhaps it would be cleaner to just have a `const std::string &shell_command argument`.

As an example, if it were truly an executable_path argument, the caller may expect spaces in the path to be valid.

tests/examples/examples_ft.cpp
```diff
@@ -215,16 +201,15 @@ TEST_F(examples_ft, transformation)
 {
     const std::string transformation_dir = TEST_TEMP_DIR;
 #ifdef _WIN32
-    const std::string delim = "\\";
-    const std::string transformation_path = "bin\\transformation_example.exe";
+    const std::string transformation_path = "bin/transformation_example.exe";
```
Brent-A: Since this is a call to a shell command, I think you can omit the .exe on windows as well, and eliminate another #if

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+    // TODO fflush needed here?
+    std::string formatted_command = shell_command_path;
+    if (!output_path.empty())
+    {
+        formatted_command += " 2>&1 > " + output_path;
+    }
+    std::cout << formatted_command << std::endl;
+    FILE *process_stream = POPEN(formatted_command.c_str(), "r");
+    ASSERT_TRUE(process_stream);
```
wes-b: This function relies on the ASSERT macro ending execution early. You should add a comment that this function should never return if it fails.

tests/sample/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
wes-b: If we rename the examples test to executable test, then this change can be undone.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,218 @@
...
+static void run_and_record_executable(const std::string &shell_command_path, const std::string &output_path)
+{
+    // TODO fflush needed here?
+    std::string formatted_command = shell_command_path;
+    if (!output_path.empty())
+    {
+        formatted_command += " 2>&1 > " + output_path;
+    }
+    std::cout << formatted_command << std::endl;
+    FILE *process_stream = POPEN(formatted_command.c_str(), "r");
```
wes-b: Can you confirm the exit code of the application?

spschul: pclose() returns the exit status, which is now returned from the function and asserted to EXIT_SUCCESS.

tests/sample/CMakeLists.txt
spschul: I'd be happy to do this but I'm not sure which is clearer. "example" is probably more clear than "sample", true, but given we have other "example"s it's still a bit confusing. Does "starter" make sense? Or something similar?

tests/examples/examples_ft.cpp
spschul: I reformatted the function and now it explicitly returns in that case, so I think this should be clear now.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,218 @@
...
+    ASSERT_TRUE(process_stream);
+    ASSERT_EQ(PCLOSE(process_stream), EXIT_SUCCESS);
+}
+
+static void test_stream_against_regexes(std::istream &input_stream, const std::vector<std::string> &regexes)
```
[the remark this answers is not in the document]
spschul: Good idea! Added.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+    auto regex_iter = regexes.cbegin();
```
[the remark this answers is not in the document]
spschul: There's a decent amount of precedent for `auto` in C++ being considered good style (https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/, though it takes him a while to really get into it) and a substantial chunk of the rest of the code in our codebase uses it extensively. It's not particularly important to me, and I'll change it, but I don't think having the whole const_iterator type spelled out would really be helpful.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,218 @@
...
+#define POPEN popen
+#define PCLOSE pclose
+#define SETENV(env, value) setenv(env, value, 1)
+#endif
+
+const static std::string TEST_TEMP_DIR = "examples_test_temp";
+
+// run the specified executable and record the output to output_path
+// if output_path is empty, just output to stdout
+static void run_and_record_executable(const std::string &shell_command_path, const std::string &output_path)
```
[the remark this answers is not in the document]
spschul: Respectfully disagreeing on this point. I don't think a developer would expect this function to modify the input, and if they were ever concerned, the function clearly uses `const` for those parameters. On the other hand, a pointer to a C++-style string- especially when C-strings are already so common- just seems confusing.

tests/examples/examples_ft.cpp
```diff
@@ -0,0 +1,58 @@
...
+    std::string fastpoint_out = "/tmp/fastpointcloud-test-out";
+    std::string fastpoint_command;
+    // TODO less ugly
+    fastpoint_command += "./fastpointcloud ";
+    fastpoint_command += fastpoint_out;
+    fastpoint_command += " 2>&1";
+#ifdef _WIN32
+    FILE *first = _popen(fastpoint_command.c_str(), "r");
+#else
+    FILE *first = popen(fastpoint_command.c_str(), "r");
```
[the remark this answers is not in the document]
spschul: Yes- they are stored to a file using the `2>&1 >` syntax later in the command.

tests/examples/examples_ft.cpp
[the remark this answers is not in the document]
wes-b: Fair enough, I do want to avoid using auto in function parameters, but you also correctly point out our code base uses it quite a bit.

tests/executables/executables_ft.cpp
```diff
@@ -0,0 +1,247 @@
...
+ * third line. It passes, so we move on to the third line and the second regex, which don't match. Note the importance
+ * of ordering.
+ */
+static void test_stream_against_regexes(std::istream &input_stream, const std::vector<std::string> &regexes)
+{
+    auto regex_iter = regexes.cbegin();
+    // ensure that all regexes are matched before the end of the file, in this order
+    while (input_stream && regex_iter != regexes.cend())
+    {
+        std::string results;
```
wes-b: How do we exit this loop if the input stream has less lines than the regexes vector?

tests/sample/CMakeLists.txt
wes-b: Not sure the context of 'starter' and what we are trying to say with that. I actually think that example and sample means the same thing here. However we have a Kinect Sample repo, and which has samples that may be incomplete or have dependencies that we don't want to bring in to the SDK project. So I like the idea of exclusively using "examples" to mean Azure-Kinect-Sensor-SDK repro and "samples" to mean the Azure Kinect Samples repo.

tests/examples/examples_ft.cpp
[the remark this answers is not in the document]
wes-b: Code gets pretty hard to follow when references are in use. And it make reviewing the code more difficult on the reviewer and thus raises the chances of bugs getting missed.  You are correct that in this instance the function uses const. To know that as a reviewer though, I have to jump to a prototype (if in an IDE), or use some Intellisense like feature to see that. However in a GitHub review I have to lose my place while reviewing to track that down and then come back. By being able to tell at the function call site if the value may be changed by the calling function means a more efficient, and ultimate effective review. So as a principle we avoid passing types by reference.

tests/sample/CMakeLists.txt
spschul: Should be fixed

tests/executables/executables_ft.cpp
spschul: Just tested to make sure, but `input_stream` is part of the `while` condition and since `getline` runs on each iteration, eventually we will reach the EOF and so `input_stream` will evaluate to `false` and the loop will exit

tests/examples/examples_ft.cpp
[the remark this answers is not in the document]
spschul: Fixed in latest commit

PR #435, opened 2019-06-24
K4A_RECORD_ENABLE_LOG_TO_A_FILE added for playback / record

### LINKED ISSUE #428: Setting `K4A_ENABLE_LOG_TO_A_FILE` causes recorder to crash
**Describe the bug**
When the `K4A_ENABLE_LOG_TO_A_FILE` environment variable is set, both k4a.dll and k4arecord.dll will try to open the same file, causing an exception to be thrown during logger creation. This exception currently isn't caught, so the entire program crashes.

**To Reproduce**
1. Set `K4A_ENABLE_LOG_TO_A_FILE` to a log file path
2. Run `k4arecorder -l 5 test.mkv`
3. The program will crash at `logger_create()` in `k4a_record_create` with a spdlog exception: "Failed opening file <path> for writing"

**Expected behavior**
The program shouldn't crash, and should log all messages to the specified file. The recorder should also operate even if creating the log file fails.

Since k4a.dll and k4arecord.dll can't open the same file and can't use the same logger instance, a separate environment variable should be created for the k4arecord logs.

 - OS with Version: Windows
 - SDK Version:  1.1.0
 - Firmware version:  Color: 1.6.98, Depth: 1.6.70

wes-b:
The logger in the SDK, when logging to a file, opens the target file exclusively. As a result when the recorder/playback attempts to log to the same file the logger crashes because it can not open the file that is already open.

The fix is to have the recorder look at a different ENV var. Going forward the recorder will use K4A_RECORD_ENABLE_LOG_TO_A_FILE and the SKD will use K4A_ENABLE_LOG_TO_A_FILE.

### I tested changes on: 
- [x] Windows
- [ ] Linux


xthexder:
It's possible for spdlog to throw an exception that makes it out to C code. We should catch this in the logger and return a proper error so that this doesn't crash the program if the path isn't found / can't be created.

wes-b:
Good point @xthexder

PR #475, opened 2019-07-10
Added global init and rwlock

### LINKED ISSUE #422: Race condition on multiple threads calling logger_register_message_callback
**Describe the bug**
If two threads simultaneously call `logger_register_message_callback`, there is a possible race where one thread will allocate g_user_logger_cb_info, and another thread will attempt to access that global before it is allocated.

Thread 1 
* Calls `logger_register_message_callback`
* increments `g_user_logger_cb_info_ref` (`count_of_registered_callbacks` is 1)
* Thread gets context switched away before completing `g_user_logger_cb_info = new …`

Thread 2
* Calls `logger_register_message_callback`
* increments `g_user_logger_cb_info_ref` (`count_of_registered_callbacks` is 2)
* Crashes dereferencing NULL during `g_user_logger_cb_info->callback`

**To Reproduce**
Get very unlucky in timing when starting a multi-threaded process that attempts to initialize logging on multiple threads at the same time.

**Expected behavior**
App doesn't crash 😊

**Logs**

**Screenshots**

 - SDK Version:  1.1.0
 - Firmware version:  N/A

**Additional context**

Brent-A:
- Adds a new reader/writer lock object
- Adds a global one time initialization helper
- Updates the logger to use the lock and init routines

 

### I tested changes on: 
- [X] Windows
- [X] Linux


include/k4ainternal/rwlock.h
```diff
@@ -0,0 +1,48 @@
...
+#include <windows.h>
+#else
+#include <pthread.h>
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct
```
wes-b: This should be hidden in the c file

include/k4ainternal/global.h
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * \param _init_function_
+ * An initialization function that returns void and takes a pointer to a _global_type_ structure.
+ * This function will be called once to initialize the global structure.
+ *
+ * \remarks
+ * This macro creates a new function with the name of the global type followed by _get(). This function
+ * returns a pointer to an initialized singleton instance of the global structure. Initialization will only
+ * happen once per module and is safe across multiple threads attempting to initialize at the same time.
+ *
```
wes-b: I might be missing something, but ```init_function``` will be called every time and needs to be robust to multiple calls.

src/global/global.c
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// This library
+#include <k4ainternal/global.h>
+
+// Dependent libraries
+#include <k4ainternal/capture.h>
+#include <azure_c_shared_utility/lock.h>
```
wes-b: do you need these includes?

tests/UnitTests/rwlock_ut/rwlock.cpp
```diff
@@ -0,0 +1,244 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <utcommon.h>
+
+#include <k4ainternal/rwlock.h>
+#include <k4ainternal/common.h>
+#include <gtest/gtest.h>
+
+#include <azure_c_shared_utility/lock.h>
```
wes-b: Here too?

src/logging/logging.cpp
```diff
@@ -71,57 +82,36 @@ k4a_result_t logger_register_message_callback(k4a_logging_message_cb_t *message_
...
-        }
 
-        delete logger_info;
-        g_user_log_level = K4A_LOG_LEVEL_OFF;
+    rwlock_acquire_write(&g_context->lock);
+
+    // The user may set or clear the callback, or set the callback to the
+    // same value it was previously. It is a failure to change the callback
+    // from an existing registration.
+    if (g_context->user_callback == NULL || message_cb == NULL || g_context->user_callback == message_cb)
```
wes-b: if (```(```g_context->user_callback == NULL ```&&``` message_cb ```!=```NULL```)``` || g_context->user_callback == message_cb)

tests/UnitTests/rwlock_ut/rwlock.cpp
```diff
@@ -0,0 +1,244 @@
...
+{
+    while (context->run_test)
+    {
+        rwlock_acquire_read(&context->test_lock);
+
+        int writer1 = context->writer1_count;
+        int writer2 = context->writer2_count;
+
+        context->reader1_count++;
+
```
MsDerekMa: Would you want a sleep here to make sure that both readers can be going in at the same time, and to make sure that the writers will block on this reader being outstanding?

src/global/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+            global.c
+            )
+
+# Consumers should #include <k4ainternal/global.h>
+target_include_directories(k4a_global PUBLIC 
+    ${K4A_PRIV_INCLUDE_DIR})
+
+# Dependencies of this library
+target_link_libraries(k4a_global PUBLIC 
+    k4ainternal::rwlock
```
MsDerekMa: Why does global depend on rwlock?

src/global/global.c
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// This library
+#include <k4ainternal/global.h>
+
+// System dependencies
+#include <windows.h>
```
MsDerekMa: Move the include Windows.h into the ifdef?

src/global/global.c
```diff
@@ -0,0 +1,45 @@
...
+#include <assert.h>
+
+#ifdef _WIN32
+
+C_ASSERT(sizeof(k4a_init_once_t) == sizeof(INIT_ONCE));
```
MsDerekMa: Add a similar assert for pthread_once_t?

tests/UnitTests/global_ut/global.cpp
```diff
@@ -0,0 +1,154 @@
...
+    int value2;
+} test_global2_t;
+
+static void globalInitFunction2(test_global2_t *global)
+{
+    ASSERT_EQ(0, global->value1);
+    ASSERT_EQ(0, global->value2);
+    ASSERT_EQ(0, g_GlobalCounter2);
+
+    g_GlobalCounter2++;
```
MsDerekMa: Should this be an interlocked increment?

include/k4ainternal/rwlock.h
Brent-A: Since the structure is different sizes on Windows and Linux, and I'd prefer to keep the lock allocation free, I think this is the cleanest we can do.

include/k4ainternal/global.h
Brent-A: This works. I've added a unit test to verify the behavior.

src/global/global.c
```diff
@@ -0,0 +1,49 @@
...
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <pthread.h>
+#endif
+#include <assert.h>
+
+#ifdef _WIN32
+
+C_ASSERT(sizeof(k4a_init_once_t) == sizeof(INIT_ONCE));
```
schultetwin1: static_assert?

```c
static_assert(sizeof(k4a_init_once_t) == sizeof(INIT_ONCE), "some message about what to do");
```

src/global/global.c
```diff
@@ -0,0 +1,49 @@
...
+
+static BOOL CALLBACK InitGlobalFunction(PINIT_ONCE InitOnce, PVOID Parameter, PVOID *lpContext)
```
schultetwin1: nit: The rest of the library uses `bool` not `BOOL`. 

```suggestion
static bool CALLBACK InitGlobalFunction(PINIT_ONCE InitOnce, PVOID Parameter, PVOID *lpContext)
```

include/k4ainternal/common.h
```diff
@@ -29,6 +29,10 @@ typedef struct _guid_t
 
 #define STRINGIFY(string) #string
 
+#ifndef C_ASSERT
+#define C_ASSERT(e) typedef char __C_ASSERT__[(e) ? 1 : -1]
```
schultetwin1: Why not just used `static_assert`?

```suggestion
```

src/global/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+add_library(k4a_global STATIC 
+            global.c
+            )
+
+# Consumers should #include <k4ainternal/global.h>
+target_include_directories(k4a_global PUBLIC 
+    ${K4A_PRIV_INCLUDE_DIR})
+
+# Dependencies of this library
+target_link_libraries(k4a_global PUBLIC
```
schultetwin1: Should this be PRIVATE instead of PUBLIC? What part of k4ainternal::rwlock do consumers of k4a_global need?

```suggestion
target_link_libraries(k4a_global PRIVATE
```

src/logging/logging.cpp
Brent-A: I think this logic is correct. You allow setting the callback (g_context->user_callback == NULL) as well as clearing the callback (message_cb !=NULL).

src/rwlock/rwlock_linux.c
```diff
@@ -0,0 +1,74 @@
...
+// This library
+#include <k4ainternal/rwlock.h>
+
+#include <k4ainternal/common.h>
+
+// System dependencies
+#include <assert.h>
+#include <pthread.h>
+
+C_ASSERT(sizeof(k4a_rwlock_t) == sizeof(pthread_rwlock_t));
```
schultetwin1: static_assert

```c
static_assert(sizeof(k4a_rwlock_t) == sizeof(pthread_rwlock_t), "some message about why we need this");
```

src/rwlock/rwlock_win32.c
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// This library
+#include <k4ainternal/rwlock.h>
+
+// System dependencies
+#include <windows.h>
+
+C_ASSERT(sizeof(k4a_rwlock_t) == sizeof(SRWLOCK));
```
schultetwin1: static_assert?

src/global/global.c
Brent-A: It's a windows.h defined function prototype, so matching their standard seems correct

src/global/global.c
Brent-A: 😢 No go in C99

PR #480, opened 2019-07-11
Updates docs to include links to depth engine readme

lauragonzalezzz:
Updates docs to include links to depth engine readme

- [ ] I [built my changes](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/blob/develop/docs/building.md) locally


docs/testing.md
```diff
@@ -1,5 +1,7 @@
 # Testing Azure Kinect SDK
 
+To run the tests, be sure the [Depth Engine](depthengine.md) is present.
```
MsDerekMa: We only need the Depth Engine for the Function and Stress tests right?  We don't need it for the Unit or Firmware tests?

lauragonzalezzz: That is correct. However, since we need the unit, functional, and firmware tests to pass in order to determine success I don't think it is necessary to distinguish, especially since I expect the tests will mostly be used to validate code changes.

tesych:
this looks good.

PR #513, opened 2019-07-16
Adds descriptive config validation error messages

### LINKED ISSUE #165: The error message is confusing when the depth mode doesn't support the fps mode the user sets
When the depth mode doesn't support the fps mode the user sets, the error message is very confusing: “depth_fps_and_mode_supported == true returned failure in validate_configuration()”. 

Changing the error message to something more human-readable will help our customers to figure out the issue easier. Here are the possible examples: 
"Depth mode X doesn’t support Y fps" or 
"The depth mode doesn’t support the FPS mode you set. "

lauragonzalezzz:
- Adds descriptive config validation error messages

### I tested changes on: 
- [x] Windows
- [ ] Linux


src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
-        config->color_format == K4A_IMAGE_FORMAT_COLOR_MJPG || config->color_format == K4A_IMAGE_FORMAT_COLOR_YUY2 ||
-        config->color_format == K4A_IMAGE_FORMAT_COLOR_NV12 || config->color_format == K4A_IMAGE_FORMAT_COLOR_BGRA32);
-    if (K4A_SUCCEEDED(result))
+    if (config->color_format != K4A_IMAGE_FORMAT_COLOR_MJPG && config->color_format != K4A_IMAGE_FORMAT_COLOR_YUY2 &&
+        config->color_format != K4A_IMAGE_FORMAT_COLOR_NV12 && config->color_format != K4A_IMAGE_FORMAT_COLOR_BGRA32)
     {
-        result = K4A_RESULT_FROM_BOOL(config->color_resolution >= K4A_COLOR_RESOLUTION_OFF &&
-                                      config->color_resolution <= K4A_COLOR_RESOLUTION_3072P);
+        result = K4A_RESULT_FAILED;
+        LOG_ERROR("The configured color format %d is not supported.", config->color_format);
```
Brent-A: Since these messages are validating that the value is within the enum range, it might make more sense to just say

LOG_ERROR("color_format is not a valid value.");
or
LOG_ERROR("%d is not a valid value of k4a_color_format_t");

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+
     if (K4A_SUCCEEDED(result))
     {
-        result = K4A_RESULT_FROM_BOOL(config->wired_sync_mode >= K4A_WIRED_SYNC_MODE_STANDALONE &&
-                                      config->wired_sync_mode <= K4A_WIRED_SYNC_MODE_SUBORDINATE);
+        if (config->camera_fps != K4A_FRAMES_PER_SECOND_5 && config->camera_fps != K4A_FRAMES_PER_SECOND_15 &&
+            config->camera_fps != K4A_FRAMES_PER_SECOND_30)
+        {
+            result = K4A_RESULT_FAILED;
+            LOG_ERROR("The configured camera fps %d is not supported.", config->camera_fps);
```
Brent-A: The previous comment is especially important here, where if they passed in a value 30 for example, they would get a message:
"The configured camera fps 30 is not supported"

Which would be confusing since K4A_FRAMES_PER_SECOND_30 is supported.

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+            if (K4A_SUCCEEDED(result))
             {
-                LOG_ERROR("Device wired_sync_mode is set to K4A_WIRED_SYNC_MODE_MASTER, so color camera must be used "
-                          "on master device. Color_resolution can not be set to K4A_COLOR_RESOLUTION_OFF.",
-                          0);
-                return K4A_RESULT_FAILED;
+                if (config->wired_sync_mode == K4A_WIRED_SYNC_MODE_SUBORDINATE && !sync_in_cable_present)
+                {
+                    result = K4A_RESULT_FAILED;
+                    LOG_ERROR("Failure to detect presence of sync in cable with wired sync mode %d.",
```
Brent-A:
```suggestion
                    LOG_ERROR("Failure to detect presence of sync in cable with wired sync mode K4A_WIRED_SYNC_MODE_SUBORDINATE.",
```

This would be a lot more readable than "Failure to detect presence of sync in cable with wired sync mode 2"

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+                              config->wired_sync_mode);
+                }
+
+                if (config->wired_sync_mode == K4A_WIRED_SYNC_MODE_MASTER)
+                {
+                    if (!sync_out_cable_present)
+                    {
+                        result = K4A_RESULT_FAILED;
+                        LOG_ERROR("Failure to detect presence of sync out cable with wired sync mode %d.",
```
Brent-A:
```suggestion
                        LOG_ERROR("Failure to detect presence of sync out cable with wired sync mode K4A_WIRED_SYNC_MODE_MASTER.",
```

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+        if (config->wired_sync_mode == K4A_WIRED_SYNC_MODE_SUBORDINATE &&
+            config->subordinate_delay_off_master_usec != 0)
         {
             uint32_t fps_in_usec = 1000000 / k4a_convert_fps_to_uint(config->camera_fps);
-            result = K4A_RESULT_FROM_BOOL(config->subordinate_delay_off_master_usec < fps_in_usec);
+            if (config->subordinate_delay_off_master_usec > fps_in_usec ||
+                config->subordinate_delay_off_master_usec < 0)
+            {
+                result = K4A_RESULT_FAILED;
+                LOG_ERROR("The configured subordinate device delay of %dusec from master device is not supported.",
```
Brent-A: Maybe split this in to two errors:
LOG_ERROR("The configured subordinate device delay of %dusec from master device cannot be negative.",...);
LOG_ERROR("The configured subordinate device delay of %dusec from master device cannot exceed one frame interval.",...);

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+            }
         }
-        else
+
+        if (config->wired_sync_mode != K4A_WIRED_SYNC_MODE_SUBORDINATE &&
+            config->subordinate_delay_off_master_usec != 0)
         {
-            result = K4A_RESULT_FROM_BOOL(config->subordinate_delay_off_master_usec == 0);
+            result = K4A_RESULT_FAILED;
+            LOG_ERROR("The configured subordinate device delay of %dusec is not supported with %d wired sync mode.",
```
Brent-A:
```suggestion
            LOG_ERROR("When wired_sync_mode is K4A_WIRED_SYNC_MODE_SUBORDINATE the subordinate_delay_off_master_usec must be 0.",
```

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+        if (depth_enabled && color_enabled)
         {
             int64_t fps = 1000000 / k4a_convert_fps_to_uint(config->camera_fps);
-            result = K4A_RESULT_FROM_BOOL((config->depth_delay_off_color_usec >= -fps) &&
-                                          (config->depth_delay_off_color_usec <= fps));
+            if (config->depth_delay_off_color_usec < -fps || config->depth_delay_off_color_usec > fps)
+            {
+                result = K4A_RESULT_FAILED;
+                LOG_ERROR(
+                    "The configured delay of %dusec between depth capture and color image capture is not supported.",
```
Brent-A:
```suggestion
                    "depth_delay_off_color_usec must be within +/- one frame interval.",
```

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
         }
-
-        if (K4A_SUCCEEDED(result))
+        else
         {
-            result = K4A_RESULT_FROM_BOOL(config->synchronized_images_only == false);
+            if (config->depth_delay_off_color_usec != 0 || config->synchronized_images_only)
+            {
+                result = K4A_RESULT_FAILED;
+                LOG_ERROR("Both depth camera and color camera must be enabled to modify configuration of depth delay "
```
Brent-A: Probably should split this in to two errors as well.
LOG_ERROR("Both depth camera and color camera must be enabled to set depth_delay_off_color_usec");
LOG_ERROR("Both depth camera and color camera must be enabled to enable synchronized_images_only");

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+                {
+                    depth_fps_and_mode_supported = true;
+                    break;
+                }
+            }
+
+            if (!depth_fps_and_mode_supported)
+            {
+                result = K4A_RESULT_FAILED;
+                LOG_ERROR("The configured depth mode %d does not support fps mode %d.",
```
Brent-A: This is tricky since the %d here will obscure the configuration. We should print the string values of the macros.

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+                {
+                    color_fps_and_res_and_format_supported = true;
+                    break;
+                }
+            }
+
+            if (!color_fps_and_res_and_format_supported)
+            {
+                result = K4A_RESULT_FAILED;
+                LOG_ERROR("The configured combination of color resolution %d, color format %d, and camera fps %d is "
```
Brent-A: Same as previous comment

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
 {
     RETURN_VALUE_IF_ARG(K4A_RESULT_FAILED, config == NULL);
     RETURN_VALUE_IF_ARG(K4A_RESULT_FAILED, device == NULL);
-    k4a_result_t result;
+    k4a_result_t result = K4A_RESULT_SUCCEEDED;
     bool depth_enabled = false;
     bool color_enabled = false;
 
-    result = K4A_RESULT_FROM_BOOL(
```
rabbitdaxi: >K4A_RESULT_FROM_BOOL [](start = 13, length = 20)

It looks good in general for the explicit logging part, but you might want to confirm with Wes about not using the K4A_RESULT_FROM_BOOL (which is not only return a k4a_result_t, but also does a TraceError). Theoretically, you can add explicit logging based on the result value without losing the trace logging as before too :)

src/sdk/k4a.c
```diff
@@ -526,171 +526,252 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+        }
     }
 
-    if (K4A_SUCCEEDED(result) && (config->wired_sync_mode == K4A_WIRED_SYNC_MODE_SUBORDINATE ||
-                                  config->wired_sync_mode == K4A_WIRED_SYNC_MODE_MASTER))
+    if (K4A_SUCCEEDED(result))
     {
-        bool sync_in_cable_present;
-        bool sync_out_cable_present;
-        result = TRACE_CALL(
```
rabbitdaxi: >TRACE_CALL [](start = 17, length = 10)

Same comment here regarding whether you can remove the TRACE_CALL

src/sdk/k4a.c
wes-b: The feedback we are getting is that tracing from the macro is confusing, so for spots in the code that we think users will commonly encounter errors, like the validation of the config structure, we a providing clearer error messages. So I am good with this, thanks @rabbitdaxi.

rabbitdaxi: Good to know :)

---
In reply to: [304560357](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/513#discussion_r304560357) [](ancestors = 304560357)

src/sdk/k4a.c
```diff
@@ -676,11 +682,19 @@ static k4a_result_t validate_configuration(k4a_context_t *device, const k4a_devi
...
+                          "depth_delay_off_color_usec.",
+                          0);
+            }
+
+            if (config->synchronized_images_only)
             {
                 result = K4A_RESULT_FAILED;
-                LOG_ERROR("Both depth camera and color camera must be enabled to modify configuration of depth delay "
-                          "or synchronized image settings.",
+                LOG_ERROR("Both depth camera and color camera must be enabled to modify configuration of "
```
wes-b: Both depth camera and color camera must be enabled to ~~modify configuration of~~ **enable**

include/k4a/k4atypes.h
```diff
@@ -280,6 +280,9 @@ typedef enum
  * \remarks
  * Binned modes reduce the captured camera resolution by combining adjacent sensor pixels into a bin.
  *
+ * \remarks
+ * If modes are added, update the k4a_depth_mode_strings array.
```
Brent-A: These remarks go in to the public API documentation. This probably doesn't belong there.

include/k4a/k4atypes.h
```diff
@@ -296,6 +299,17 @@ typedef enum
     K4A_DEPTH_MODE_PASSIVE_IR,     /**< Passive IR only, captured at 1024x1024. */
 } k4a_depth_mode_t;
 
+/** Depth sensor capture mode strings.
+ *
+ * \remarks
+ * If strings are added, update the k4a_depth_mode_t enum.
+ */
+static const char *k4a_depth_mode_strings[6] = {
```
Brent-A: I'm not sure about adding this to the public API surface. This may be more appropriate as an internal helper.

There's no way to easily know the bounds of a value that is being used to lookup in to this array, so it would be pretty easy for someone to translate k4a_depth_modes_strings[modevariable] without verifying that modevariable is between 0 and 5.

It might be better to have an internal helper function like this:

```C
#define K4A_DEPTH_MODE_TO_STRING_CASE( mode ) case mode: return #mode
const char* k4a_depth_mode_to_string(k4a_depth_mode_t mode)
{
   switch (mode)
   {
       K4A_DEPTH_MODE_TO_STRING_CASE( K4A_DEPTH_MODE_OFF );
       K4A_DEPTH_MODE_TO_STRING_CASE( K4A_DEPTH_MODE_NFOV_UNBINNED );
       K4A_DEPTH_MODE_TO_STRING_CASE( K4A_DEPTH_MODE_WFOV_UNBINNED );
       etc.
   }
   return "unexpected k4a_depth_mode_t value";
}
```

That function will ensure that the inputs are sanitized, and that the strings always match the enum names or it will be a build break. I think with high warning levels, clang may provide a build break if a new enum value is added and not all the switch cases are covered.

include/k4a/k4atypes.h
```diff
@@ -280,15 +280,13 @@ typedef enum
...
- * \remarks
- * If modes are added, update the k4a_depth_mode_strings array.
- *
  * \xmlonly
  * <requirements>
  *   <requirement name="Header">k4atypes.h (include k4a/k4a.h)</requirement>
  * </requirements>
  * \endxmlonly
  */
+// Be sure to update k4a_depth_mode_to_string in k4a.c if enum values are added.
```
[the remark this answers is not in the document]
lauragonzalezzz: I did have that initially before pushing my changes but Brent recommended pulling this comment out of the doxygen block entirely. Suggestion @Brent-A ?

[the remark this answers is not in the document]
Brent-A: /internal is mainly used if we want to produce a second set of doxygen output explaining how to use internal components. For this sort of comment, which isn't about the usage of the API but more about how to make edits to the file, I'm not sure that it needs to go in any doxygen docs. Since we aren't building docs with /internal, it doesn't really matter.

PR #515, opened 2019-07-17
Fix exposure settings2

### LINKED ISSUE #362: k4a_image_get_exposure_usec returns different value than exposure set with k4a_device_set_color_control
**Describe the bug**
When setting an exposure value using k4a_device_set_color_control, I expect that exposure value to be the same as the one used by in the color image later which I can examine via k4a_image_get_exposure_usec. However, they do not always match.

**To Reproduce**
1. Call k4a_device_set_color_control with an exposure value and manual exposure control:
k4a_device_set_color_control(deviceHandle, K4A_COLOR_CONTROL_EXPOSURE_TIME_ABSOLUTE, K4A_COLOR_CONTROL_MODE_MANUAL, exposureValue);

2. Get the exposure value again to verify that it was set correctly with the right value in manual mode:
k4a_device_get_color_control(deviceHandle, K4A_COLOR_CONTROL_EXPOSURE_TIME_ABSOLUTE, &currentExposureSetting, &currentExposureValue);

3. Start the cameras with k4a_device_start_cameras

4. Get a capture with k4a_device_get_capture and get the color frame with k4a_capture_get_color_image.

5. Get exposure setting from the image with k4a_image_get_exposure_usec and compare it to the one set before with k4a_device_set_color_control.

**Expected behavior**
The value returned by k4a_image_get_exposure_usec should be the same as the value set by k4a_device_set_color_control before starting the camera.

**Logs**
[outputLogs.txt](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/files/3204281/outputLogs.txt)

**Screenshots**
Call to k4a_device_set_color_control and output of k4a_device_get_color_control
exposureSet
k4a_image_get_exposure_usec outputted in filename of color image 
exposureSeen

 - OS with Version: Windows 10 Version 1809 (Build 17763.437)
 - SDK Version: 1.0.2
 - Firmware version: 1.6.98

**Additional context**
In the situation we're experiencing the bug in a 5 Eden setup, as can be seen in the logs. We set the exposure for all the devices to 31250 us, and then the devices have varying exposure values that do not match this setting of 40000 and 41670.

wes-b:
- Adds exposure mapping table implemented by firmware to Windows & Linux so they can map as best as possible the exposure settings.
- Updated color_ft tests to verify both 50Hz and 60Hz implementations. We only testing while the sensor is off. We are not testing the runtime ability to change these values.
- Throughput_perf was updated to show the exposure being set, the value that was actually set, and the value that is returned on each camera image.

### I tested changes on: 
- [x] Windows
- [x] Linux

I also run through the different exposure and power line settings with throughput_perf


src/color/uvc_camerareader.cpp
```diff
@@ -1289,4 +1293,49 @@ UVCCameraReader::DecodeMJPEGtoBGRA32(uint8_t *in_buf, const size_t in_size, uint
...
+
+// Returns exposure in 100us time base
+uint32_t UVCCameraReader::MapK4aExposureToLinux(uint32_t K4aExposure_usec)
+{
+    // We map to the expected exposure then convert to 100us time base to ensure we roll over to the next exposure
+    // setting in the same way Windows does.
+    for (uint32_t x = 0; x < COUNTOF(device_exposure_mapping); x++)
+    {
+        uint32_t mapped_exposure = device_exposure_mapping[x].exposure_mapped_50Hz_usec;
+        ;
```
cdedmonds: Intentional?

wes-b: Nope :)

PR #545, opened 2019-07-25
Add the current module path to the search path

### LINKED ISSUE #538: Include SDK path in DLL search path when attempting to load the depth engine
**Describe the bug**
When the Azure Kinect Sensor SDK attempts to load the Depth Engine is just uses load library. In cases where the executing code is not in the same location as the Azure Kinect Sensor SDK, the depth engine won't be found. Similar to #310, we should add the local path to the search on Windows. This should then match the .NET and Linux search paths.

**To Reproduce**
1. Create a project in Unity with the Azure Kinect Sensor SDK
2. Run the project to start streaming sensor data
3. Note that the start fails because the Depth Engine wasn't loaded

**Expected behavior**
The depth engine should be able to be loaded in 

 - SDK Version:  Latest Develop

MsDerekMa:
- Adding logic to add the current module path to the search path when attempting to load the depth engine.

### I tested changes on: 
- [X] Windows
- [ ] Linux

### DIFF
--- a/src/dynlib/dynlib_windows.c
+++ b/src/dynlib/dynlib_windows.c
@@ -126,7 +126,7 @@ k4a_result_t dynlib_create(const char *name, uint32_t major_ver, uint32_t minor_
         }
     }
 
-    if (dllDirectory != 0)
+    if (dllDirectory != NULL)
     {
         if (RemoveDllDirectory(dllDirectory) == 0)
         {
--- a/src/dynlib/dynlib_windows.c
+++ b/src/dynlib/dynlib_windows.c
@@ -68,6 +68,8 @@ static DLL_DIRECTORY_COOKIE add_current_module_to_search()
 
     fileName[0] = L'\0';
 
+    // This adds the directory of the current module (k4a.dll) to the loader's path.
+    // This will mimic how C# loads DLLs. The loader for C code only loads from the path of the current executable
     DLL_DIRECTORY_COOKIE dllDirectory = AddDllDirectory(path);
     if (dllDirectory == 0)
     {
--- a/src/dynlib/dynlib_windows.c
+++ b/src/dynlib/dynlib_windows.c
@@ -66,7 +66,7 @@ static DLL_DIRECTORY_COOKIE add_current_module_to_search()
         return NULL;
     }
 
-    fileName[0] = '\0';
+    fileName[0] = L'\0';
 
     DLL_DIRECTORY_COOKIE dllDirectory = AddDllDirectory(path);
     if (dllDirectory == 0)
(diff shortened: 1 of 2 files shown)


src/dynlib/dynlib_windows.c
```diff
@@ -37,6 +37,46 @@ static char *generate_file_name(const char *name, uint32_t major_ver, uint32_t m
...
+        LOG_WARNING("Failed to get current module file name (%d).", GetLastError());
+        return NULL;
+    }
+
+    // GetModuleFileName give the full path, but AddDllDirectory requires a directory path. Remove
+    // the filename portion of this path. This should only be running in the "K4A.dll" assembly,
+    // but verify this assumption before truncating the path. If this code is used from a different
+    // file, then this needs to be modified to search for the path separators.
+    size_t length = wcslen(path);
+    wchar_t *fileName = path + (length - 7);
```
wes-b: Can you define "k4a.dll" once and use sizeof or countof to get the '7' you need here?

src/dynlib/dynlib_windows.c
```diff
@@ -37,6 +37,46 @@ static char *generate_file_name(const char *name, uint32_t major_ver, uint32_t m
...
+        return NULL;
+    }
+
+    if (GetModuleFileNameW(hModule, path, sizeof(path)) == 0)
+    {
+        LOG_WARNING("Failed to get current module file name (%d).", GetLastError());
+        return NULL;
+    }
+
+    // GetModuleFileName give the full path, but AddDllDirectory requires a directory path. Remove
```
schultetwin1: What if we used [PathCchRemoveFileSpec](https://docs.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchremovefilespec) instead of manually stripping off k4a.dll? That way we don't need to worry what the module name is.

src/dynlib/dynlib_windows.c
```diff
@@ -37,6 +37,46 @@ static char *generate_file_name(const char *name, uint32_t major_ver, uint32_t m
...
+    wchar_t *fileName = path + (length - 7);
+    if (fileName <= path || _wcsicmp(fileName, L"k4a.DLL") != 0)
+    {
+        LOG_WARNING("The file name of the current module is not expected.", 0);
+        return NULL;
+    }
+
+    fileName[0] = L'\0';
+
+    DLL_DIRECTORY_COOKIE dllDirectory = AddDllDirectory(path);
```
schultetwin1: Can you add a comment here explaining why we need to do this.

```suggestion
    // This adds the directory of the current module (k4a.dll) to the loader's path.
    // This will mimic how C# loads DLLs. The loader for C code only loads from the path of the current executable
    DLL_DIRECTORY_COOKIE dllDirectory = AddDllDirectory(path);
```

src/dynlib/dynlib_windows.c
```diff
@@ -80,6 +124,14 @@ k4a_result_t dynlib_create(const char *name, uint32_t major_ver, uint32_t minor_
         }
     }
 
+    if (dllDirectory != 0)
```
schultetwin1: nit: NULL

Or switch `add_current_module_to_search` to return 0.

```suggestion
    if (dllDirectory != NULL)
```

src/dynlib/dynlib_windows.c
MsDerekMa: I'll use Matt's suggestion as it removes the need for assuming the file name.

src/dynlib/dynlib_windows.c
MsDerekMa: Sounds good.

PR #546, opened 2019-07-25
Update Build Infrastructure to support multiple devices.

wes-b:
Fixes: #614 

This:
- Removes the need for the connection in the lab
- Issues a reset to the devices in the lab instead of power cycling
- Modifies the AzureKinectFirmwareTool to perform operations on all devices in the lab.


azure-pipelines.yml
```diff
@@ -749,6 +759,13 @@ jobs:
       continueOnError: true
 
     # The binary is built from https://microsoft.visualstudio.com/Analog/_git/systems.rustyhmdkit
-    - script: '$(System.ArtifactsDirectory)/findconnectedport/linux/findconnectedport'
-      workingDirectory: '$(System.ArtifactsDirectory)/findconnectedport/linux/'
+#    - script: '$(System.ArtifactsDirectory)/findconnectedport/linux/findconnectedport'
+#      workingDirectory: '$(System.ArtifactsDirectory)/findconnectedport/linux/'
+#      displayName: 'Reset K4A Device'
+
+    - script: './x86_64-linux-clang-relwithdebinfo/bin/AzureKinectFirmwareTool -r'
```
MsDerekMa: Why do we have a second reset after all of the tests have completed?

wes-b: The connex is actually powering off the hardware, so I assumed that at a min it was at least resetting the HW.

tools/updater/main.c
```diff
@@ -934,174 +862,220 @@ static k4a_result_t command_inspect_firmware(updater_command_info_t *command_inf
...
+        result = command_query_device(&sub_command_info);
+        if (!K4A_SUCCEEDED(result))
+        {
+            return result;
+        }
 
-    // Wait until the update operation is complete and query the device to get status...
-    bool update_failed = K4A_FAILED(wait_update_operation_complete(command_info->firmware_handle, &finalStatus));
+        // Load and parse the firmware file information...
+        result = command_inspect_firmware(&sub_command_info);
```
MsDerekMa: This loads and inspects the firmware. This should get called outside the for loop to prevent leaking memory and in general to prevent duplicating loading from disk.

tools/updater/main.c
```diff
@@ -934,174 +862,220 @@ static k4a_result_t command_inspect_firmware(updater_command_info_t *command_inf
...
-    if (!K4A_SUCCEEDED(result))
-    {
-        return result;
-    }
+        // Make a copy of the command so that only 1 device is updated at a time.
+        updater_command_info_t sub_command_info = { 0 };
+        sub_command_info.device_count = 1;
+        sub_command_info.device_serial_number[0] = command_info->device_serial_number[device_index];
+        sub_command_info.firmware_handle = command_info->firmware_handle;
+        sub_command_info.firmware_serial_number = command_info->firmware_serial_number;
```
MsDerekMa: I might clear command_info->firmware_handle and firmware_serial_number as sub_command_info now own both.

tools/updater/main.c
```diff
@@ -1152,10 +1126,13 @@ int main(int argc, char **argv)
...
     close_all_handles(&command_info);
 
-    if (command_info.device_serial_number)
+    for (uint32_t device_index = 0; device_index < command_info.device_count; device_index++)
     {
-        free(command_info.device_serial_number);
-        command_info.device_serial_number = NULL;
+        if (command_info.device_serial_number[device_index])
+        {
+            free(command_info.device_serial_number[device_index]);
```
MsDerekMa: Technically this should be firmware_free_serial_number.

tools/updater/main.c
```diff
@@ -934,174 +862,220 @@ static k4a_result_t command_inspect_firmware(updater_command_info_t *command_inf
...
-    if (!K4A_SUCCEEDED(result))
-    {
-        return result;
-    }
+    k4a_result_t finalCmdStatus = K4A_RESULT_SUCCEEDED;
+    k4a_result_t result;
 
-    result = firmware_reset_device(command_info->firmware_handle);
-    if (!K4A_SUCCEEDED(result))
+    for (uint32_t device_index = 0; device_index < command_info->device_count; device_index++)
```
MsDerekMa: Agreed, I think we should just serialize the calls here so that we complete the reset before we go on to the next one.

azure-pipelines.yml
wes-b: It is removed now

src/firmware/firmware.c
```diff
@@ -156,6 +205,11 @@ void firmware_destroy(firmware_t firmware_handle)
         depthmcu_destroy(firmware->depthmcu);
         firmware->depthmcu = NULL;
     }
+    if (firmware->colormcu)
+    {
+        colormcu_destroy(firmware->colormcu);
+        firmware->colormcu = NULL;
+    }
```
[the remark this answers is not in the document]
wes-b: Good catch!

src/firmware/firmware.c
```diff
@@ -227,6 +289,7 @@ k4a_buffer_result_t firmware_get_device_serialnum(firmware_t firmware_handle,
     RETURN_VALUE_IF_ARG(K4A_BUFFER_RESULT_FAILED, serial_number_size == NULL);
 
     firmware_context_t *firmware = firmware_t_get_context(firmware_handle);
+    RETURN_VALUE_IF_ARG(K4A_BUFFER_RESULT_FAILED, firmware->depthmcu == NULL);
 
     return TRACE_BUFFER_CALL(depthmcu_get_serialnum(firmware->depthmcu, serial_number, serial_number_size));
```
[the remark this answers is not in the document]
wes-b: good question, this function can be removed. There is no reason to have a get serial number function if we open the module with a serial number.

src/firmware/firmware.c
```diff
@@ -386,3 +450,67 @@ k4a_result_t parse_firmware_package(const uint8_t *firmware_buffer,
...
+    {
+        b_result = colormcu_get_serialnum(color, NULL, &serial_number_length);
+    }
+    else if (depth)
+    {
+        b_result = depthmcu_get_serialnum(depth, NULL, &serial_number_length);
+    }
+    else
+    {
+        LOG_ERROR("No Color or Depth handle provided\n", 0);
```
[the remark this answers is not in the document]
wes-b: I thought the firmware module was exclusively using LOG_* and the code in the updater folder was using printf?

tools/updater/main.c
```diff
@@ -91,99 +91,130 @@ static void print_supprted_commands()
...
     {
-        if (K4A_RESULT_SUCCEEDED != depthmcu_create(deviceIndex, &device))
+        char firstCharacter = argv[next][0];
+        if (firstCharacter != '-' && firstCharacter != '/')
         {
-            printf("ERROR: %d: Failed to open device\n", deviceIndex);
-            continue;
+            // If the user is passing in a serial number, then we use it and exit. We will later check for it
+            device_serial_number = argv[next];
         }
```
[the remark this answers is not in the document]
wes-b: I did this in an earlier iteration, but we need to validate the target serial number is present and fail out here rather than let ensure_firmware_open attempt to open it the first time as it has retry logic and will take ~30s to faile.

src/firmware/firmware.c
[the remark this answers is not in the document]
wes-b: As per our chat, I will review and remove as needed

tools/updater/main.c
```diff
@@ -707,407 +669,345 @@ static k4a_result_t ensure_firmware_open(updater_command_info_t *command_info)
...
+    result = load_firmware_file(firmware_info);
     if (!K4A_SUCCEEDED(result))
     {
         return result;
     }
 
-    parse_firmware_package(command_info->firmware_buffer,
-                           command_info->firmware_size,
-                           &command_info->firmware_package_info);
+    parse_firmware_package(firmware_info->buffer, firmware_info->size, firmware_info);
```
MsDerekMa: Since the buffer is in firmware_info now, should this get updated to just take firmware_info?

tools/updater/main.c
```diff
@@ -707,407 +669,345 @@ static k4a_result_t ensure_firmware_open(updater_command_info_t *command_info)
...
-    {
-        printf("ERROR: Failed to get current versions\n\n");
-        return K4A_RESULT_FAILED;
     }
 
     return K4A_RESULT_SUCCEEDED;
 }
 
-static k4a_result_t command_inspect_firmware(updater_command_info_t *command_info)
+static k4a_result_t command_inspect_firmware(char *firmware_path, firmware_package_info_t *firmware_info)
```
MsDerekMa: Since it is a little confusing, could we rename this to "command_load_and_inspect_firmware"?

tools/updater/main.c
```diff
@@ -707,407 +669,345 @@ static k4a_result_t ensure_firmware_open(updater_command_info_t *command_info)
...
 static k4a_result_t command_update_device(updater_command_info_t *command_info)
 {
-    firmware_status_summary_t finalStatus;
+    firmware_status_summary_t finalFwCommandStatus;
+    k4a_result_t finalCmdStatus = K4A_RESULT_SUCCEEDED;
 
-    k4a_result_t result = ensure_firmware_open(command_info);
+    // Load and parse the firmware file information...
+    firmware_package_info_t firmware_info;
+    k4a_result_t result = command_inspect_firmware(command_info->firmware_path, &firmware_info);
```
MsDerekMa: I'm not seeing where you are freeing firmware_info.

tests/FirmwareTests/firmware_fw.cpp
```diff
@@ -55,13 +55,6 @@ class firmware_fw : public ::testing::Test
             firmware_handle = nullptr;
         }
 
-        if (serial_number != nullptr)
```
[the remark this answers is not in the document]
wes-b: The serial number is read once for all tests. So if it does change, all the remaining tests will fail. I am ok with that.

Issue #495, opened 2019-07-14
Please clarify Depth Engine Linux OpenGL requirements in docs


jcsoo:
The current docs in [Using Azure Kinect SDK](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/blob/develop/docs/usage.md) are vague on Linux dependencies, simply listing "OpenGL". The page [Azure Kinect sensor SDK system requirements](https://docs.microsoft.com/en-us/azure/Kinect-dk/system-requirements) on the main site could be a bit more specific - for Linux it lists "Linux Ubuntu 18.04 (x64) with OpenGLv4.4 or later GPU driver", but it's not clear if there is some subset of SDK functionality that works with a lower OpenGL version or if nothing works at all.

In fact it appears that the depth sensor capture code in the Depth Engine is specifically what has the dependency on OpenGL 4.4. The enumeration and firmware examples (including k4aviewer when only streaming from the color camera) all seem to run fine on OpenGL 3.3, but k4arecord fails with this error:

```
Device version: Rel; C: 1.6.98; D: 1.6.70[6109.7]; A: 1.6.14
[2019-07-14 06:45:44.871] [trace] [t=27389] ../src/sdk/k4a.c (704): k4a_device_start_cameras(). k4a_device_start_cameras starting
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (722): k4a_device_start_cameras(). Starting camera's with the following config.
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (723): k4a_device_start_cameras().     color_format:0
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (724): k4a_device_start_cameras().     color_resolution:2
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (725): k4a_device_start_cameras().     depth_mode:2
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (726): k4a_device_start_cameras().     camera_fps:2
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (727): k4a_device_start_cameras().     synchronized_images_only:0
[2019-07-14 06:45:44.871] [info] [t=27389] ../src/sdk/k4a.c (728): k4a_device_start_cameras().     depth_delay_off_color_usec:0
[...]

PR #566, opened 2019-07-30
Transformation custom image support

### LINKED ISSUE #260: Support mapping image other than the depth map from depth space to color space
The instance map in the K4ABT SDK is a map in depth space. But currently there is no way for them to map the instance map from the depth space to color space. It would be useful to have an API to help map the image other than the depth map from depth space to color space.

christianmakela:
- This change implements support for custom image transformation using depth engine version 2.0.0.
-
-

### I tested changes on: 
- [x] Windows
- [ ] Linux

Added new test cases for transformation.cpp.


include/k4a/k4a.h
```diff
@@ -1985,6 +1985,88 @@ K4A_EXPORT k4a_result_t k4a_transformation_depth_image_to_color_camera(k4a_trans
...
+/** Transforms depth map and a custom image into the geometry of the color camera.
+ *
+ * \param transformation_handle
+ * Transformation handle.
+ *
+ * \param depth_image
+ * Handle to input depth image.
+ *
+ * \param custom_image
+ * Handle to input custom image.
```
wes-b: In the resolution of the depth image, right? If we expect certain key properties set for this image, we should be clear about it here.

include/k4a/k4a.h
```diff
@@ -1985,6 +1985,88 @@ K4A_EXPORT k4a_result_t k4a_transformation_depth_image_to_color_camera(k4a_trans
...
+ * or k4a_image_create_from_buffer().
+ *
+ * \remarks
+ * Using linear interpolation could create new values to \p transformed_custom_image which do no exist in \p
+ * custom_image. Setting \p use_linear_interpolation to false will prevent this from happenning but will result in less
+ * smooth image.
+ *
+ * \returns
+ * ::K4A_RESULT_SUCCEEDED if \p transformed_depth_image and \p transformed_custom_image were successfully written and
+ * ::K4A_RESULT_FAILED otherwise.
```
wes-b: If we return failed, are we ensuring all resources we may have allocated have been freed?

include/k4a/k4atypes.h
```diff
@@ -420,6 +442,23 @@ typedef enum
...
+ * <requirements>
+ *   <requirement name="Header">k4atypes.h (include k4a/k4a.h)</requirement>
+ * </requirements>
+ * \endxmlonly
+ */
+typedef enum
+{
+    K4A_TRANSFORMATION_INTERPOLATION_TYPE_NEAREST = 0, /**< Nearest neighbor interpolation */
+    K4A_TRANSFORMATION_INTERPOLATION_TYPE_LINEAR,      /**< Linear interpolation */
+} k4a_transformation_interpolation_type_t;
```
wes-b: will deloader_transform_engine_process_frame ever get called where this parameter will be a don't care? If so we might want to add XXX_NONE.

include/k4ainternal/k4aplugin.h
```diff
@@ -87,11 +87,21 @@ typedef enum
  */
 typedef enum
 {
-    K4A_TRANSFORM_ENGINE_TYPE_COLOR_TO_DEPTH = 0, /**< Transforms the color image into the geometry of the depth camera
-                                                   */
-    K4A_TRANSFORM_ENGINE_TYPE_DEPTH_TO_COLOR /**< Transforms the depth image into the geometry of the color camera */
+    K4A_TRANSFORM_ENGINE_TYPE_COLOR_TO_DEPTH = 0,   /**< Transforms color image into the geometry of the depth camera */
+    K4A_TRANSFORM_ENGINE_TYPE_DEPTH_TO_COLOR,       /**< Transforms depth image into the geometry of the color camera */
+    K4A_TRANSFORM_ENGINE_TYPE_DEPTH_AUX8_TO_COLOR,  /**< Transforms depth + 8 bit aux into the color camera geometry */
```
wes-b: Here are use the term AUX8 and AUX16, but the image type is MONO8 and MONO16. I think we should align these terms or at least tie them together with a comment.

src/dewrapper/dewrapper.c
```diff
@@ -258,7 +262,14 @@ static int depth_engine_thread(void *param)
                                                     &outputCaptureInfo,
                                                     NULL);
             tickcounter_get_current_ms(dewrapper->tick, &stop_time);
-            if (deresult != K4A_DEPTH_ENGINE_RESULT_SUCCEEDED)
+            if (deresult == K4A_DEPTH_ENGINE_RESULT_FATAL_ERROR_WAIT_PROCESSING_COMPLETE_FAILED ||
+                deresult == K4A_DEPTH_ENGINE_RESULT_FATAL_ERROR_GPU_TIMEOUT)
+            {
+                LOG_ERROR("Timeout during depth engine process frame. This could indicate GPU is hung.", 0);
```
wes-b: FYI, this is an implementation detail in the public space. you might just want to say the GPU is in a bad state, and we must restart.

include/k4a/k4a.h
```diff
@@ -1985,6 +1985,88 @@ K4A_EXPORT k4a_result_t k4a_transformation_depth_image_to_color_camera(k4a_trans
...
+ * \xmlonly
+ * <requirements>
+ *   <requirement name="Header">k4a.h (include k4a/k4a.h)</requirement>
+ *   <requirement name="Library">k4a.lib</requirement>
+ *   <requirement name="DLL">k4a.dll</requirement>
+ * </requirements>
+ * \endxmlonly
+ */
+K4A_EXPORT k4a_result_t
+k4a_transformation_custom_depth_image_to_color_camera(k4a_transformation_t transformation_handle,
```
rabbitdaxi: >k4a_transformation_custom_depth_image_to_color_camera [](start = 0, length = 53)

nit: not sure whether "k4a_transformation_depth_custom_image_to_color_camera" is a better name, because custom_depth_image feels like a customized depth image

include/k4a/k4a.hpp
```diff
@@ -845,6 +845,31 @@ class transformation
         }
     }
 
+    /** Transforms depth map and an custom image into the geometry of the color camera.
```
rabbitdaxi: >an [](start = 33, length = 2)

a

include/k4a/k4a.hpp
```diff
@@ -845,6 +845,31 @@ class transformation
...
+     * Throws error on failure
+     *
+     * \sa k4a_transformation_custom_depth_image_to_color_camera
+     */
+    void custom_depth_image_to_color_camera(const image &depth_image,
```
rabbitdaxi: >custom_depth_image_to_color_camera [](start = 9, length = 34)

nit: may be change to depth_custom_image_to_color_camera

src/dewrapper/dewrapper.c
```diff
@@ -152,6 +152,10 @@ static k4a_result_t depth_engine_start_helper(dewrapper_context_t *dewrapper,
         if (deresult != K4A_DEPTH_ENGINE_RESULT_SUCCEEDED)
         {
             LOG_ERROR("Depth engine create and initialize failed with error code: %d.", deresult);
+            if (deresult == K4A_DEPTH_ENGINE_RESULT_FATAL_ERROR_GPU_OPENGL_CONTEXT)
+            {
+                LOG_ERROR("OpenGL context creation failed. You could try updating your graphics drivers.", 0);
```
rabbitdaxi: do we need to be more specific on the opengl version minimum requirement in this message?

src/sdk/k4a.c
```diff
@@ -1018,15 +1019,58 @@ k4a_result_t k4a_transformation_depth_image_to_color_camera(k4a_transformation_t
...
+    return TRACE_CALL(transformation_custom_depth_image_to_color_camera(transformation_handle,
+                                                                        depth_image_buffer,
+                                                                        &depth_image_descriptor,
+                                                                        custom_image_buffer,
+                                                                        &dummy_descriptor,
+                                                                        transformed_depth_image_buffer,
+                                                                        &transformed_depth_image_descriptor,
+                                                                        transformed_custom_image_buffer,
+                                                                        &dummy_descriptor,
+                                                                        K4A_TRANSFORMATION_INTERPOLATION_TYPE_LINEAR,
```
rabbitdaxi: may be add a comment to specify these two parameter in this function are noop?

src/transformation/rgbz.c
```diff
@@ -556,9 +798,28 @@ k4a_buffer_result_t transformation_depth_image_to_color_camera_internal(
...
+    }
+    else if (custom_image_descriptor->format == K4A_IMAGE_FORMAT_MONO16)
+    {
+        context.enable_custom16 = true;
+    }
+
+    context.interpolation_type = interpolation_type;
+    context.invalid_value = (uint16_t)(invalid_custom_value & 0xffff);
+
     if (K4A_FAILED(TRACE_CALL(transformation_depth_to_color(&context))))
```
rabbitdaxi: >transformation_depth_to_color [](start = 30, length = 29)

is this function should be named with the depth+custom similarly as other places?

rabbitdaxi:
You might want to edit the PR description that is also fixes #495

include/k4a/k4a.h
christianmakela: I can add a remark about that.

include/k4a/k4a.h
christianmakela: This function doesn't actually do any resource allocation. It's no different from k4a_transformation_depth_image_to_color_camera or k4a_transformation_color_image_to_depth_camera. All resources will be allocated by k4a_transformation_create.

include/k4a/k4a.hpp
christianmakela: Fixed.

include/k4a/k4a.hpp
christianmakela: I will change this to mimic whatever name is defined in k4a.h.

include/k4a/k4atypes.h
christianmakela: So I originally had a "don't care" option called INTERPOLATION_TYPE_ANY here. It was removed after the first review since it was decided that the client of k4a should always select an interpolation method. You're right that deloader_transform_engine_process_frame could in specific situations use the "don't care" option but that's an internal detail that we probably don't want to expose to the client of k4a. What I could do is add back the "don't care" option to the k4aplugin.h header - but then it's not in sync with the k4a.h enum. @mbleyer Do you have an opinion about this?

include/k4ainternal/k4aplugin.h
christianmakela: I can rename these to CUSTOM8 and CUSTOM16. On a related note there was a request to rename MONO8 and MONO16 to CUSTOM8 and CUSTOM16 in k4atypes.h.

src/dewrapper/dewrapper.c
christianmakela: Yeah, I know. I tried to make the log message helpful without going too much into the details. Do you have suggestion on how to change this?

src/dewrapper/dewrapper.c
christianmakela: I thought about that but then realized if we change the version in depth engine dll then this message will be out of sync. So I don't know, should we define the version here?

src/sdk/k4a.c
christianmakela: Yeah, I can add a comment.

wes-b:
I would just report the GPU has encountered an unrecoverable error and the SDK needs to be restartes

include/k4ainternal/k4aplugin.h
wes-b: ok

include/k4a/k4a.h
wes-b: thanks

src/transformation/rgbz.c
christianmakela: I didn't rename the internal functions since there is only one implementation of depth_to_color that handles both cases; depth only and depth + custom.

src/dewrapper/dewrapper.c
christianmakela: I'll add the version.

include/k4a/k4atypes.h
```diff
@@ -412,6 +412,28 @@ typedef enum
...
+    /** Single channel image type MONO16.
+     *
+     * \details
+     * Each pixel of MONO16 is a single channel two bytes of little endian unsigned data.
+     *
+     * \details
+     * Stride indicates the length of each line in bytes and should be used to determine the start location of each
+     * line of the image in memory.
+     */
+    K4A_IMAGE_FORMAT_MONO16,
```
cdedmonds: >K4A_IMAGE_FORMAT_MONO16 [](start = 4, length = 23)

Did we agree to change these to custom?

christianmakela: Yes.

src/dewrapper/dewrapper.c
christianmakela: I changed the error message to your suggestion.

include/k4a/k4atypes.h
christianmakela: I will leave this as is for now.

src/dewrapper/dewrapper.c
rabbitdaxi: How about: "OpenGL context creation failed. Requires OpenGLv4.4 or above. You could try updating your graphics drivers.

---
In reply to: [309440703](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/566#discussion_r309440703) [](ancestors = 309440703)

PR #592, opened 2019-08-05
Convert all playback timestamps to use device time

### LINKED ISSUE #590: IMU timestamps are incorrect in recording files.
**Describe the bug**

The timestamps returned by the playback API are offset by `record_configuration.start_timestamp_offset_usec`, which is inconsistent with the rest of the playback API.
The seek timestamp functionality will also select the wrong IMU sample in most cases because of this offset.

**To Reproduce**
See recently added test in `feature/support-custom-track-record-playback`: [open_start_offset_file](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/246/files#diff-cf969dd6e43c1a7237f7fb864c562a87R794)

**Expected behavior**

Timestamps should be consistent across the playback API. Currently this means timestamps should be relative to the start of the recording.

**Logs**
```
[ RUN      ] playback_ut.open_start_offset_file
[2019-08-05 10:39:28.003] [error] [t=40204] ..\tests\RecordTests\UnitTest\test_helpers.cpp (268): validate_imu_sample(). PlaybackTest, imu_sample.acc_timestamp_usec is incorrect. Actual: 1001150, Expected: 1150
..\tests\RecordTests\UnitTest\playback_ut.cpp(869): error: Value of: validate_imu_sample(imu_sample, imu_timestamp)
  Actual: false
Expected: true
[  FAILED  ] playback_ut.open_start_offset_file (43 ms)
```

 - OS with Version: All
 - SDK Version:  develop branch
 - Firmware version:  N/A

xthexder:
- Changes `get_next_capture()` / `get_previous_capture()` to return images with device timestamp instead of relative file timestamps.
- Deprecate `k4a_playback_get_last_timestamp_usec()` and replace with `k4a_playback_get_recording_length_usec()`
- Add `K4A_PLAYBACK_SEEK_DEVICE_TIME` for seeking to a device timestamp

### I tested changes on: 
- [x] Windows
- [ ] Linux

I've split this off from #246 so it's easier to review, and included some of the test suite changes from the other PR.

### DIFF
--- a/include/k4arecord/playback.hpp
+++ b/include/k4arecord/playback.hpp
@@ -288,11 +288,11 @@ public:
 
     /** Get the last valid timestamp in the recording
      *
-     * \sa k4a_playback_get_last_timestamp_usec
+     * \sa k4a_playback_get_recording_length_usec
      */
-    std::chrono::microseconds get_last_timestamp() const noexcept
+    std::chrono::microseconds get_recording_length() const noexcept
     {
-        return std::chrono::microseconds(k4a_playback_get_last_timestamp_usec(m_handle));
+        return std::chrono::microseconds(k4a_playback_get_recording_length_usec(m_handle));
     }
 
     /** Set the image format that color captures will be converted to. By default the conversion format will be the same
--- a/src/record/sdk/playback.cpp
+++ b/src/record/sdk/playback.cpp
@@ -338,14 +338,14 @@ k4a_result_t k4a_playback_seek_timestamp(k4a_playback_t playback_handle,
     if (origin == K4A_PLAYBACK_SEEK_END)
     {
         uint64_t offset_ns = (uint64_t)(-offset_usec * 1000);
-        if (offset_ns > context->last_timestamp_ns)
+        if (offset_ns > context->last_file_timestamp_ns)
         {
             // If the target timestamp is negative, clamp to 0 so we don't underflow.
             target_time_ns = 0;
         }
         else
         {
-            target_time_ns = context->last_timestamp_ns + 1 - offset_ns;
+            target_time_ns = context->last_file_timestamp_ns + 1 - offset_ns;
         }
     }
     else
@@ -379,7 +379,7 @@ uint64_t k4a_playback_get_recording_length_usec(k4a_playback_t playback_handle)
 
     k4a_playback_context_t *context = k4a_playback_t_get_context(playback_handle);
     RETURN_VALUE_IF_ARG(0, context == NULL);
-    return context->last_timestamp_ns / 1000;
+    return context->last_file_timestamp_ns / 1000;
 }
 
 uint64_t k4a_playback_get_last_timestamp_usec(k4a_playback_t playback_handle)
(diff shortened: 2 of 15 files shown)


include/k4arecord/playback.hpp
```diff
@@ -288,11 +288,11 @@ class playback
 
     /** Get the last valid timestamp in the recording
      *
-     * \sa k4a_playback_get_last_timestamp_usec
+     * \sa k4a_playback_get_recording_length_usec
      */
-    std::chrono::microseconds get_last_timestamp() const noexcept
+    std::chrono::microseconds get_recording_length() const noexcept
```
xthexder: The C++ API for playback has not been released yet, so this change should be fine without worrying about deprecation.

src/record/sdk/playback.cpp
```diff
@@ -364,6 +373,15 @@ k4a_result_t k4a_playback_seek_timestamp(k4a_playback_t playback_handle,
     return K4A_RESULT_SUCCEEDED;
 }
 
+uint64_t k4a_playback_get_recording_length_usec(k4a_playback_t playback_handle)
+{
+    RETURN_VALUE_IF_HANDLE_INVALID(0, k4a_playback_t, playback_handle);
+
+    k4a_playback_context_t *context = k4a_playback_t_get_context(playback_handle);
+    RETURN_VALUE_IF_ARG(0, context == NULL);
+    return context->last_timestamp_ns / 1000;
```
wes-b: now that there is file time stamp and device timestamp, it is not clear to me what last_timestamp_ns refers to.

xthexder:
The "Public K4A Symbols" step seems to be failing randomly, but all the tests have passed. I'll try running the tests again.

tools/k4aviewer/k4arecordingdockcontrol.cpp
```diff
@@ -392,7 +385,7 @@ bool K4ARecordingDockControl::PlaybackThreadFn(PlaybackThreadState *state)
         {
             if (image)
             {
-                image.set_timestamp(image.get_device_timestamp() + state->TimestampOffset);
+                image.set_timestamp(image.get_device_timestamp());
             }
```
wes-b: Does the viewer now even care to get and track TimestampOffset?

xthexder: The start offset is still displayed as part of the recording info, but it isn't needed for playback at all.

src/record/sdk/playback.cpp
xthexder: All timestamps internal to the playback API (like this one), are still relative to start of file. This could probably be refactored to reflect this now that we have 2 time domains, but for now I think that can be left for another PR, since it's not visible to users of the API.

tools/k4aviewer/k4arecordingdockcontrol.cpp
wes-b: Right, so it seems like dead code in the viewer

src/record/sdk/playback.cpp
wes-b: can you just rename last_timestamp_ns to something like last_file_timestamp_ns?

tools/k4aviewer/k4arecordingdockcontrol.cpp
xthexder: I've removed all references to `state->TimestampOffset` already. `m_startTimestampOffsetUsec` was left in so that the value can still be displayed.

wes-b:
> The "Publish K4A Symbols" step seems to be failing randomly, but all the tests have passed. I'll try running the tests again.

It was down, hopefully it stays up now that things are succeeding.

src/record/sdk/playback.cpp
xthexder: Done

PR #594, opened 2019-08-05
Made image_create format aware

### LINKED ISSUE #587: Incorrect image size calculation in k4a_image_create function for K4A_IMAGE_FORMAT_COLOR_NV12 format
Image size calculation inside `k4a_image_create` function is incorrect in case of `K4A_IMAGE_FORMAT_COLOR_NV12` format:

```c
k4a_image_create(K4A_IMAGE_FORMAT_COLOR_NV12, 1280, 720, 1280, &image);
```
creates image with buffer size `921600` bytes (I guess it is calculated as `sride * height` which gives `1280 * 720 = 921600` in our case).

But it is not correct size for NV12 image `1280x720`. Correct size must be `3 * stride * height / 2`  that is `1382400` bytes.

You can say that this is by design and I have to specify different stride for such NV12 image. But let's start color streaming from device in `K4A_IMAGE_FORMAT_COLOR_NV12` format and `K4A_COLOR_RESOLUTION_720P` resolution, then wait for the first capture and ask color image from it using `k4a_capture_get_color_image`. As a result we'll get image with the following parameters:
* format: ` K4A_IMAGE_FORMAT_COLOR_NV12`
* width: `1280` pixels
* height: `720` pixels
* stride: `1280` bytes (exactly as in our call to `k4a_image_create`)
* size: `1382400` bytes (but `k4a_image_create` returns image with size = `921600` bytes)

That is, at least we have some inconsistency between `k4a_image_create` and `k4a_capture_get_color_image` function for `K4A_IMAGE_FORMAT_COLOR_NV12` images.

SDK Version:  1.1.1

Brent-A:
- Made image_create stride aware

image_create will now validate the basic requirements of known formats, and allocate sufficient memory for the NV12 format.

### I tested changes on: 
- [X] Windows
- [X] Linux


src/image/image.c
```diff
@@ -171,10 +171,140 @@ k4a_result_t image_create(k4a_image_format_t format,
...
+        }
+        break;
+    }
+
+    // 1 Byte per pixel
+    case K4A_IMAGE_FORMAT_CUSTOM8:
+    {
+        if (stride_bytes < 1 * width_pixels)
+        {
+            LOG_ERROR("Insufficient stride (%d bytes) to represent image width (%d pixels).",
```
wes-b: It would be nice if these error messages included the format too.

src/image/image.c
```diff
@@ -171,10 +171,140 @@ k4a_result_t image_create(k4a_image_format_t format,
...
+        {
+            size = (size_t)height_pixels * (size_t)stride_bytes;
+            result = K4A_RESULT_SUCCEEDED;
+        }
+        break;
+    }
+
+    // Unknown
+    case K4A_IMAGE_FORMAT_CUSTOM:
+    default:
```
wes-b: Perhaps we should not support this type. I suppose we already have code though relying on this

PR #596, opened 2019-08-06
Add overload functions that returns k4a::image in k4a::transformation

### LINKED ISSUE #582: [Proposal] Create output k4a::image in k4a::transformation
I propose that create output memory in k4a::transformation.
It is no necessary to write code to create output image by developer.
Therefor, It will shortens the developer side code.

We can implement it with little changes in k4a.hpp.

```patch
     /** Transforms the depth map into the geometry of the color camera.
      * Throws error on failure
      *
      * \sa k4a_transformation_depth_image_to_color_camera
      */
-    void depth_image_to_color_camera(const image &depth_image, image *transformed_depth_image) const
+    void depth_image_to_color_camera(const image &depth_image, 
+                                     const image &color_image,
+                                     image *transformed_depth_image) const
     {
+        if (transformed_depth_image)
+        {
+            *transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
+                                                     color_image.get_width_pixels(),
+                                                     color_image.get_height_pixels(),
+                                                     color_image.get_width_pixels() * static_cast<int32_t>(sizeof(uint16_t)));
+        }

         k4a_result_t result = k4a_transformation_depth_image_to_color_camera(m_handle,
                                                                              depth_image.handle(),
                                                                              transformed_depth_image->handle());
         if (K4A_RESULT_SUCCEEDED != result)
         {
             throw error("Failed to convert depth map to color camera geometry!");
         }
     }

     /** Transforms the color image into the geometry of the depth camera.
      * Throws error on failure
      *
      * \sa k4a_transformation_color_image_to_depth_camera
      */
     void color_image_to_depth_camera(const image &depth_image,
                                      const image &color_image,
                                      image *transformed_color_image) const
     {
+        if (transformed_color_image)
+        {
+            *transformed_color_image = image::create(K4A_IMAGE_FORMAT_COLOR_BGRA32,
+                                                     depth_image.get_width_pixels(),
+                                                     depth_image.get_height_pixels(),
+                                                     depth_image.get_width_pixels() * 4 * static_cast<int32_t>(sizeof(uint8_t)));
+        }

         k4a_result_t result = k4a_transformation_color_image_to_depth_camera(m_handle,
                                                                              depth_image.handle(),
                                                                              color_image.handle(),
                                                                              transformed_color_image->handle());
         if (K4A_RESULT_SUCCEEDED != result)
         {
             throw error("Failed to convert color image to depth camera geometry!");
         }
     }

     /** Transforms the depth image into 3 planar images representing X, Y and Z-coordinates of corresponding 3d points.
      * Throws error on failure.
      *
      * \sa k4a_transformation_depth_image_to_point_cloud
      */
     void depth_image_to_point_cloud(const image &depth_image, k4a_calibration_type_t camera, image *xyz_image) const
     {
+        if (xyz_image)
+        {
+            *xyz_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
+                                       depth_image.get_width_pixels(),
+                                       depth_image.get_height_pixels(),
+                                       depth_image.get_width_pixels() * 3 * static_cast<int32_t>(sizeof(int16_t)));
+        }

         k4a_result_t result =
             k4a_transformation_depth_image_to_point_cloud(m_handle, depth_image.handle(), camera, xyz_image->handle());
         if (K4A_RESULT_SUCCEEDED != result)
         {
             throw error("Failed to transform depth image to point cloud!");
         }
     }

```

I think that is no problem in general use case. It make simple.
And, I hope that will helpful for developers by simplifying code they write.
```cpp
// Before
k4a::image transformed_depth_image = k4a::image::create(K4A_IMAGE_FORMAT_DEPTH16, color_image.get_width_pixels(), color_image.get_height_pixels(), color_image.get_width_pixels() * static_cast<int32_t>(sizeof(uint16_t)));
transformation.depth_image_to_color_camera(depth_image, &transformed_depth_image);

k4a::image transformed_color_image = k4a::image::create(K4A_IMAGE_FORMAT_COLOR_BGRA32, depth_image.get_width_pixels(), depth_image.get_height_pixels(), depth_image.get_width_pixels() * 4 * static_cast<int32_t>(sizeof(uint8_t)));
transformation.color_image_to_depth_camera( depth_image, color_image, &transformed_color_image );

k4a::image xyz_image = image::create(K4A_IMAGE_FORMAT_CUSTOM, depth_image.get_width_pixels(), depth_image.get_height_pixels(), depth_image.get_width_pixels() * 3 * static_cast<int32_t>(sizeof(int16_t)));
transformation.depth_image_to_point_cloud(depth_image, K4A_CALIBRATION_TYPE_DEPTH, &xyz_image);
```
```cpp
// After
k4a::image transformed_depth_image;
transformation.depth_image_to_color_camera(depth_image, color_image, &transformed_depth_image);

k4a::image transformed_color_image;
transformation.color_image_to_depth_camera(depth_image, color_image, &transformed_color_image);

k4a::image xyz_image;
transformation.depth_image_to_point_cloud(depth_image, K4A_CALIBRATION_TYPE_DEPTH, &xyz_image);
```

What do you think? Thanks,

UnaNancyOwen:
This pull request will implements #582.
That proposal is add following two features related <code>k4a::transformation</code>.
(There are already implemented similar features in C# API.)

* [x] Creates output <code>k4a::image</code> in <code>k4a::transformation</code>.
* [x] Add overload functions that returns <code>k4a::image</code>.

I think this change will helpfull developers by simplifies user code. It make simple.
(It doesn't break existing user code by this change. It can be using the legacy user code as it is.)
```cpp
k4a::image transformed_color_image = transformation.color_image_to_depth_camera(depth_image, color_image);
```

- Creates output <code>k4a::image</code> in <code>k4a::transformation</code>.
- Add overload functions that returns <code>k4a::image</code>.
  ```patch
    void image_to_camera(const image &input_image, image *output_image) const
    {
        ...
    }

  + image image_to_camera(const image &input_image) const
  + {
  +     output_image = image::create( ... );
  +     ...
  +     return output_image;
  + }
  ```

### I tested changes on: 
- [x] Windows
- [ ] Linux

### DIFF
--- a/include/k4a/k4a.hpp
+++ b/include/k4a/k4a.hpp
@@ -847,15 +847,6 @@ public:
      */
     void depth_image_to_color_camera(const image &depth_image, image *transformed_depth_image) const
     {
-        if (transformed_depth_image)
-        {
-            *transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
-                                                     m_color_calibration.resolution_width,
-                                                     m_color_calibration.resolution_height,
-                                                     m_color_calibration.resolution_width *
-                                                         static_cast<int32_t>(sizeof(uint16_t)));
-        }
-
         k4a_result_t result = k4a_transformation_depth_image_to_color_camera(m_handle,
                                                                              depth_image.handle(),
                                                                              transformed_depth_image->handle());
@@ -873,7 +864,11 @@ public:
      */
     image depth_image_to_color_camera(const image &depth_image) const
     {
-        image transformed_depth_image;
+        image transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
+                                                      m_color_calibration.resolution_width,
+                                                      m_color_calibration.resolution_height,
+                                                      m_color_calibration.resolution_width *
+                                                          static_cast<int32_t>(sizeof(uint16_t)));
         depth_image_to_color_camera(depth_image, &transformed_depth_image);
         return transformed_depth_image;
     }
@@ -882,6 +877,7 @@ public:
      * Throws error on failure
      *
      * \sa k4a_transformation_depth_image_to_color_camera_custom
+     * Transforms the output in to the existing caller provided \p transformed_depth_image \p transformed_custom_image.
      */
     void depth_image_to_color_camera_custom(const image &depth_image,
                                             const image &custom_image,
@@ -890,24 +886,6 @@ public:
                                             k4a_transformation_interpolation_type_t interpolation_type,
                                             uint32_t invalid_custom_value) const
     {
-        if (transformed_depth_image)
-        {
-            *transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
-                                                     m_color_calibration.resolution_width,
-                                                     m_color_calibration.resolution_height,
-                                                     m_color_calibration.resolution_width *
-                                                         static_cast<int32_t>(sizeof(uint16_t)));
-        }
-
-        if (transformed_custom_image)
-        {
-            *transformed_custom_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
-                                                      m_color_calibration.resolution_width,
-                                                      m_color_calibration.resolution_height,
-                                                      m_color_calibration.resolution_width * 3 *
-                                                          static_cast<int32_t>(sizeof(int16_t)));
-        }
-
         k4a_result_t result = k4a_transformation_depth_image_to_color_camera_custom(m_handle,
                                                                                     depth_image.handle(),
                                                                                     custom_image.handle(),
@@ -925,6 +903,7 @@ public:
      * Throws error on failure
      *
      * \sa k4a_transformation_depth_image_to_color_camera_custom
+     * Creates a new image with the output.
      */
     std::pair<image, image>
     depth_image_to_color_camera_custom(const image &depth_image,
@@ -932,8 +911,16 @@ public:
                                        k4a_transformation_interpolation_type_t interpolation_type,
                                        uint32_t invalid_custom_value) const
     {
-        image transformed_depth_image;
-        image transformed_custom_image;
+        image transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
+                                                      m_color_calibration.resolution_width,
+                                                      m_color_calibration.resolution_height,
+                                                      m_color_calibration.resolution_width *
+                                                          static_cast<int32_t>(sizeof(uint16_t)));
+        image transformed_custom_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
+                                                       m_color_calibration.resolution_width,
+                                                       m_color_calibration.resolution_height,
+                                                       m_color_calibration.resolution_width * 3 *
+                                                           static_cast<int32_t>(sizeof(int16_t)));
         depth_image_to_color_camera_custom(depth_image,
                                            custom_image,
                                            &transformed_depth_image,
@@ -947,20 +934,12 @@ public:
      * Throws error on failure
      *
      * \sa k4a_transformation_color_image_to_depth_camera
+     * Transforms the output in to the existing caller provided \p transformed_color_image.
      */
     void color_image_to_depth_camera(const image &depth_image,
                                      const image &color_image,
                                      image *transformed_color_image) const
     {
-        if (transformed_color_image)
-        {
-            *transformed_color_image = image::create(K4A_IMAGE_FORMAT_COLOR_BGRA32,
-                                                     m_depth_calibration.resolution_width,
-                                                     m_depth_calibration.resolution_height,
-                                                     m_depth_calibration.resolution_width * 4 *
-                                                         static_cast<int32_t>(sizeof(uint8_t)));
-        }
-
         k4a_result_t result = k4a_transformation_color_image_to_depth_camera(m_handle,
                                                                              depth_image.handle(),
                                                                              color_image.handle(),
@@ -975,10 +954,15 @@ public:
      * Throws error on failure
      *
      * \sa k4a_transformation_color_image_to_depth_camera
+     * Creates a new image with the output.
      */
     image color_image_to_depth_camera(const image &depth_image, const image &color_image) const
     {
-        image transformed_color_image;
+        image transformed_color_image = image::create(K4A_IMAGE_FORMAT_COLOR_BGRA32,
+                                                      m_depth_calibration.resolution_width,
+                                                      m_depth_calibration.resolution_height,
+                                                      m_depth_calibration.resolution_width * 4 *
+                                                          static_cast<int32_t>(sizeof(uint8_t)));
         color_image_to_depth_camera(depth_image, color_image, &transformed_color_image);
         return transformed_color_image;
     }
@@ -987,18 +971,10 @@ public:
      * Throws error on failure.
      *
      * \sa k4a_transformation_depth_image_to_point_cloud
+     * Transforms the output in to the existing caller provided \p xyz_image.
      */
     void depth_image_to_point_cloud(const image &depth_image, k4a_calibration_type_t camera, image *xyz_image) const
     {
-        if (xyz_image)
-        {
-            *xyz_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
-                                       m_depth_calibration.resolution_width,
-                                       m_depth_calibration.resolution_height,
-                                       m_depth_calibration.resolution_width * 3 *
-                                           static_cast<int32_t>(sizeof(int16_t)));
-        }
-
         k4a_result_t result =
             k4a_transformation_depth_image_to_point_cloud(m_handle, depth_image.handle(), camera, xyz_image->handle());
         if (K4A_RESULT_SUCCEEDED != result)
@@ -1011,10 +987,15 @@ public:
      * Throws error on failure.
      *
      * \sa k4a_transformation_depth_image_to_point_cloud
+     * Creates a new image with the output.
      */
     image depth_image_to_point_cloud(const image &depth_image, k4a_calibration_type_t camera) const
     {
-        image xyz_image;
+        image xyz_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
+                                        m_depth_calibration.resolution_width,
+                                        m_depth_calibration.resolution_height,
+                                        m_depth_calibration.resolution_width * 3 *
+                                            static_cast<int32_t>(sizeof(int16_t)));
         depth_image_to_point_cloud(depth_image, camera, &xyz_image);
         return xyz_image;
     }
--- a/include/k4a/k4a.hpp
+++ b/include/k4a/k4a.hpp
@@ -869,6 +869,7 @@ public:
      * Throws error on failure
      *
      * \sa k4a_transformation_depth_image_to_color_camera
+     * Creates a new image with the output.
      */
     image depth_image_to_color_camera(const image &depth_image) const
     {
--- a/include/k4a/k4a.hpp
+++ b/include/k4a/k4a.hpp
@@ -918,11 +918,22 @@ public:
                                                       m_color_resolution.height,
                                                       m_color_resolution.width *
                                                           static_cast<int32_t>(sizeof(uint16_t)));
+        int32_t bytes_per_pixel;
+        switch (custom_image.get_format())
+        {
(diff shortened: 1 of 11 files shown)


include/k4a/k4a.hpp
```diff
@@ -836,6 +842,14 @@ class transformation
      */
     void depth_image_to_color_camera(const image &depth_image, image *transformed_depth_image) const
     {
+        if (transformed_depth_image)
```
Brent-A: We need to retain the ability to transform in to an existing output image. This enables callers to re-use the output image during a processing loop and not need to allocate each time. I think with this version, the image passed in by the caller will be lost and a new image will be allocated.

I think we should remove the creation of the new image in this overload, and only have it create the image in the new overload. This should make it more clear that in one version you are transforming to an existing image, and in the other you are creating a new one.

@williampMSFT, what do you think?

include/k4a/k4a.hpp
```diff
@@ -836,6 +842,14 @@ class transformation
      */
```
Brent-A:
```suggestion
     * Transforms the output in to the existing caller provided \p transformed_depth_image.
     */
```

include/k4a/k4a.hpp
```diff
@@ -845,6 +859,18 @@ class transformation
         }
     }
 
+    /** Transforms the depth map into the geometry of the color camera.
+     * Throws error on failure
+     *
+     * \sa k4a_transformation_depth_image_to_color_camera
+     */
```
Brent-A:
```suggestion
     * Creates a new image with the output.
     */
```

include/k4a/k4a.hpp
williampMSFT: I agree - it makes it much clearer that one is allocating memory on your behalf and the other isn't, which is relevant if you're trying to tune perf.

---
In reply to: [311126657](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/596#discussion_r311126657) [](ancestors = 311126657)

UnaNancyOwen: I'll fix it. Thanks,

include/k4a/k4a.hpp
```diff
@@ -904,8 +1005,22 @@ class transformation
...
+    {
+        image xyz_image;
+        depth_image_to_point_cloud(depth_image, camera, &xyz_image);
+        return xyz_image;
+    }
+
 private:
     k4a_transformation_t m_handle;
+    k4a_calibration_camera_t m_color_calibration;
+    k4a_calibration_camera_t m_depth_calibration;
```
williampMSFT: These are relatively large structs - they include camera intrinsics and extrinsics and some other stuff, none of which we seem to actually be using in this class.

As far as I can tell, we're only retaining these to keep track of the image dimensions - consider just using image dimensions

UnaNancyOwen: I got it. Thanks,

include/k4a/k4a.hpp
```diff
@@ -870,6 +921,28 @@ class transformation
...
+    {
+        image transformed_depth_image;
+        image transformed_custom_image;
+        depth_image_to_color_camera_custom(depth_image,
+                                           custom_image,
+                                           &transformed_depth_image,
+                                           &transformed_custom_image,
+                                           interpolation_type,
+                                           invalid_custom_value);
+        return { transformed_depth_image, transformed_custom_image };
```
williampMSFT: >transformed_depth_image, transformed_custom_image  [](start = 17, length = 50)

nit: I'm pretty sure this results in the copy constructors getting invoked to make the std::pair, which ends up doing an addref/decref.
Consider wrapping these in std::move to dodge the refcount ++--.

UnaNancyOwen: Is it <code>{std::move(transformed_depth_image), std::move(transformed_custom_image)}</code>?

williampMSFT: Yep!

---
In reply to: [311196260](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/596#discussion_r311196260) [](ancestors = 311196260)

UnaNancyOwen: Thanks,

include/k4a/k4a.hpp
UnaNancyOwen: @williampMSFT Which do you think is better?
```cpp
private:
    k4a_transformation_t m_handle;
    std::pair<int32_t, int32_t> m_color_resolution;
    std::pair<int32_t, int32_t> m_depth_resolution;
};
```
```cpp
private:
    k4a_transformation_t m_handle;
    struct resolution {
        int32_t widht;
        int32_t height;
    };
    resolution m_color_resolution;
    resolution m_depth_resolution;
};
```

williampMSFT: I personally prefer the second one, but I'll leave it up to you :)

UnaNancyOwen: I implemented with the second one. 👍

include/k4a/k4a.hpp
```diff
@@ -870,10 +901,42 @@ class transformation
...
+                                       const image &custom_image,
+                                       k4a_transformation_interpolation_type_t interpolation_type,
+                                       uint32_t invalid_custom_value) const
+    {
+        image transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
+                                                      m_color_resolution.width,
+                                                      m_color_resolution.height,
+                                                      m_color_resolution.width *
+                                                          static_cast<int32_t>(sizeof(uint16_t)));
+        image transformed_custom_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
```
Brent-A: This looks like it's actually incorrect. Check out docs for [k4a_transformation_depth_image_to_color_camera_custom](https://microsoft.github.io/Azure-Kinect-Sensor-SDK/develop/group___functions_gac00dd00e7612a86382e3d0a130f276bb.html#gac00dd00e7612a86382e3d0a130f276bb).

transformed_custom_image must be of CUSTOM8 or CUSTOM16. and the output image needs to be of the matching format (with a stride that's appropriate for the pixel width).

@rabbitdaxi added those functions and should probably review the usage here.

include/k4a/k4a.hpp
```diff
@@ -904,8 +985,31 @@ class transformation
...
 
+    /** Transforms the depth image into 3 planar images representing X, Y and Z-coordinates of corresponding 3d points.
+     * Throws error on failure.
+     *
+     * \sa k4a_transformation_depth_image_to_point_cloud
+     * Creates a new image with the output.
+     */
+    image depth_image_to_point_cloud(const image &depth_image, k4a_calibration_type_t camera) const
+    {
+        image xyz_image = image::create(K4A_IMAGE_FORMAT_CUSTOM,
```
Brent-A: The resolution of the xyz image needs to match the [resolution of the depth image](https://microsoft.github.io/Azure-Kinect-Sensor-SDK/develop/group___functions_ga7385eb4beb9d8892e8a88cf4feb3be70.html#ga7385eb4beb9d8892e8a88cf4feb3be70) passed in. 

One use case of this function is for the depth image to be transformed to the perspective of the color camera first, then converted to a point cloud. This allows you to align color values with the point cloud points.

UnaNancyOwen: Oops, I got it all wrong. Thanks.

include/k4a/k4a.hpp
UnaNancyOwen: How is this?
```cpp
image transformed_custom_image = image::create(custom_image.get_format(),
                                               m_color_resolution.width,
                                               m_color_resolution.height,
                                               m_color_resolution.width * (custom_image.get_stride_bytes() /
                                                                           custom_image.get_width_pixels()));
```

UnaNancyOwen: I still don't understand how many channels of custom image requires in <code>k4a_transformation_depth_image_to_color_camera_custom()</code>.
Is it single channel?

Brent-A: I believe it is only single channel. @christianmakela, is that correct?

cdedmonds: Single channel is correct. It can be either 8 or 16 bits per pixel.

rabbitdaxi: Yes, it is single channel, and because it can be 8 or 16bits per pixel, therefore, recommend to based on the custom_image to determine the byteperpixel for the transformed_custom_image

---
In reply to: [311326510](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/596#discussion_r311326510) [](ancestors = 311326510)

rabbitdaxi: >K4A_IMAGE_FORMAT_CUSTOM [](start = 55, length = 23)

this should be the custom_image's format, you can use the image_get_format to retrieve it from the original custom image

rabbitdaxi: This is good

---
In reply to: [311296317](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/596#discussion_r311296317) [](ancestors = 311296317)

UnaNancyOwen: I got it. Thanks!

UnaNancyOwen: Thank you for the advice.

include/k4a/k4a.hpp
```diff
@@ -870,10 +901,42 @@ class transformation
...
+    {
+        image transformed_depth_image = image::create(K4A_IMAGE_FORMAT_DEPTH16,
+                                                      m_color_resolution.width,
+                                                      m_color_resolution.height,
+                                                      m_color_resolution.width *
+                                                          static_cast<int32_t>(sizeof(uint16_t)));
+        image transformed_custom_image = image::create(custom_image.get_format(),
+                                                       m_color_resolution.width,
+                                                       m_color_resolution.height,
+                                                       m_color_resolution.width * (custom_image.get_stride_bytes() /
```
Brent-A: stride / width doesn't necessarily give you the pixel size. Stride can be larger than the width * pixels size to account for alignment concerns. or to subsection an image buffer. (a DEPTH16 image could be 5 pixels wide and have a stride of 16 bytes for example)

The only way to know the correct size would be to look at the pixel format.

UnaNancyOwen: How is this? Are there any good other methods for that?
```cpp
const int32_t bytes_per_pixel = static_cast<int32_t>(
            (custom_image.get_format() == K4A_IMAGE_FORMAT_CUSTOM8) ? sizeof(int8_t) : sizeof(int16_t));
image transformed_custom_image = image::create(custom_image.get_format(),
                                               m_color_resolution.width,
                                               m_color_resolution.height,
                                               m_color_resolution.width * bytes_per_pixel);
```

Brent-A: I think is ok, since the transformation function should fail if it is anything other than CUSTOM8 or CUSTOM16. The only risk is that if there was a future CUSTOM32 or some other type that the transformation function allows you to transform this code will default to 2 bytes per pixel which may be wrong. It should still be safe though, the worst that would happen is the function would fail or you would allocate extra memory.

In general I prefer to be more explicit and validate that custom_image.get_format() is one of the two exact types expected and fail otherwise. In this case you are adding logic that is only valid for those two formats out of a larger set of possibilities, so its safer to validate that the image is one of those two formats first rather then rely on later logic to ensure correctness.

UnaNancyOwen: I see. It may be redundant, but I think this will work and able to prepare for future expansion.
```cpp
int32_t bytes_per_pixel;
switch (custom_image.get_format())
{
    case K4A_IMAGE_FORMAT_CUSTOM8:
        bytes_per_pixel = static_cast<int32_t>(sizeof(int8_t));
        break;
    case K4A_IMAGE_FORMAT_CUSTOM16:
        bytes_per_pixel = static_cast<int32_t>(sizeof(int16_t));
        break;
    default:
        throw error("Failed to support this format of custom image!");
}
image transformed_custom_image = image::create(custom_image.get_format(),
                                               m_color_resolution.width,
                                               m_color_resolution.height,
                                               m_color_resolution.width * bytes_per_pixel);
```

Brent-A: I like it.

Maybe it would benefit us to add a k4a_get_pixel_size_from_image_format function for this sort of thing in the future, but for now this seems pretty clean.

UnaNancyOwen: > Maybe it would benefit us to add a k4a_get_pixel_size_from_image_format function for this sort of thing in the future

I agree.

Brent-A:
Thank you @UnaNancyOwen!

Issue #32, opened 2019-02-06
Add support for source indexing


schultetwin1:
Getting started here:
https://docs.microsoft.com/en-us/azure/devops/artifacts/symbols/setting-up-github-sourcelinking?view=azure-devops

schultetwin1:
[Source linking](https://github.com/dotnet/sourcelink) is a relatively new and so not so well documented alternative to source indexing. It's safer than source indexing since it does not rely on the debugger running arbitrary scripts and is better documented than source indexing.

Source linking has been in Visual Studio since VS2017 but a bug means its not usable till VS2019. See [here](https://github.com/dotnet/sourcelink#prerequisites-for-c-projects) for more info.

As a result, the source linking in this repo will not be useful until we start releasing VS2019 binaries.

PR #634, opened 2019-08-12
Enables SourceLink for Windows Builds

schultetwin1:

Turns on source linking for VS2019 builds

Source linking is a new way to index sources in PDBs. It embeded a JSON into the PDB which tells the debugger where to find the source files. Note, this only works in VS2019 because of file case issues in VS2017. Therefore issue #32 will not be resolved until we start shipping binaries from VS2019.

### I tested changes on:
- [ ] Windows
- [ ] Linux

To manually test this, I built a PDB with these flags off the develop branch. Then I deleted all my sources and tried to debug a unit test in VS2019. VS2019 was able to download the sources from GitHub.

Note: Source linking does not appear to work in WinDBG. I have an internal ask out about that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ project(K4A LANGUAGES C CXX
 
 option(K4A_BUILD_DOCS "Build K4A doxygen documentation" OFF)
 option(K4A_MTE_VERSION "Skip FW version check" OFF)
+option(K4A_SOURCE_LINK "Enable source linking on MSVC" OFF)
 
 include(GitCommands)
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -55,10 +56,30 @@ endif()
...
+
+    # Turn off incremental linking
     include(MSVCLinkerFlags)
+
+    # Enable source linking
+    # NOTE: Dependencies are not properly setup here.
+    #       Currently, CMake does not know to re-link if SOURCE_LINK_JSON changes
+    #       Currently, CMake does not re-generate SOURCE_LINK_JSON if git's HEAD changes
+    if (K4A_SOURCE_LINK)
+        if ("${CMAKE_C_COMPILER_VERSION}" VERSION_GREATER_EQUAL "19.20")
```
MsDerekMa: Do we have plans around when we are going to switch to 2019? And when we'll no longer build on 2017?

schultetwin1: These are two different asks.

1. When are we going to release binaries built with VS2019 instead of binaries built with VS2017?
   This we can start doing soon. VS2017 and VS2019 have toolchains which (should) produce ABI compatible binaries. Therefore, we can release binaries built by VS2019 and VS2017 will still work.

2. When are we going to stop building VS2017?
    I imagine this will not be for awhile.

PR #645, opened 2019-08-14
Fix warnings on clang 8 / GCC 9

xthexder:
- Ignore some new warnings added in Clang 8
- Fix warnings from Clang 8 / GCC 9
- Updated ImGui to v1.72b to fix some warnings

### I tested changes on: 
- [x] Windows
- [x] Linux

Added comments inline about specific warning fixes.

### DIFF
--- a/cmake/k4aCompilerFlags.cmake
+++ b/cmake/k4aCompilerFlags.cmake
@@ -29,6 +29,11 @@ if ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
         # Added in clang 5
         list(APPEND CLANG_ALL_WARNINGS "-Wno-zero-as-null-pointer-constant") # Allow zero as nullptr
     endif()
+    if (NOT (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS "8.0.0"))
+        # Added in clang 8
+        list(APPEND CLANG_ALL_WARNINGS "-Wno-extra-semi-stmt") # Allow semi-colons to be used after #define's
+        list(APPEND CLANG_ALL_WARNINGS "-Wno-atomic-implicit-seq-cst") # Allow use of __sync_add_and_fetch() atomic
+    endif()
     set(CLANG_WARNINGS_AS_ERRORS "-Werror")
     add_compile_options(${CLANG_ALL_WARNINGS})
     add_compile_options(${CLANG_WARNINGS_AS_ERRORS})
--- a/include/k4a/k4a.hpp
+++ b/include/k4a/k4a.hpp
@@ -730,12 +730,12 @@ struct calibration : public k4a_calibration_t
      */
     static calibration get_from_raw(char *raw_calibration,
                                     size_t raw_calibration_size,
-                                    k4a_depth_mode_t depth_mode,
-                                    k4a_color_resolution_t color_resolution)
+                                    k4a_depth_mode_t target_depth_mode,
+                                    k4a_color_resolution_t target_color_resolution)
     {
         calibration calib;
         k4a_result_t result =
-            k4a_calibration_get_from_raw(raw_calibration, raw_calibration_size, depth_mode, color_resolution, &calib);
+            k4a_calibration_get_from_raw(raw_calibration, raw_calibration_size, target_depth_mode, target_color_resolution, &calib);
 
         if (K4A_RESULT_SUCCEEDED != result)
         {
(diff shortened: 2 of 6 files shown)


include/k4a/k4a.hpp
```diff
@@ -730,12 +730,12 @@ struct calibration : public k4a_calibration_t
      */
     static calibration get_from_raw(char *raw_calibration,
                                     size_t raw_calibration_size,
-                                    k4a_depth_mode_t depth_mode,
-                                    k4a_color_resolution_t color_resolution)
+                                    k4a_depth_mode_t target_depth_mode,
```
xthexder: `depth_mode` shadows a field from the parent class `k4a_calibration_t`

include/k4ainternal/global.h
```diff
@@ -41,7 +47,7 @@ void global_init_once(k4a_init_once_t *init_once, k4a_init_once_function_t *init
     static _global_type_ _##_global_type_##_private;                                                                   \
     static void fn_##_global_type_##_init_function(void)                                                               \
     {                                                                                                                  \
-        memset(&_##_global_type_##_private, 0, sizeof(_##_global_type_##_private));                                    \
+        DEFAULT_INIT(_global_type_, _##_global_type_##_private);                                                       \
```
xthexder: `K4A_DECLARE_GLOBAL` is used with `logger_global_context_t`, which contains C++ types, and should not be memset.

Warning: `clearing an object of type ‘struct logger_global_context_t’ with no trivial copy-assignment; use assignment or value-initialization instead`

src/record/internal/matroska_read.cpp
```diff
@@ -193,7 +193,7 @@ k4a_result_t parse_mkv(k4a_playback_context_t *context)
                     if (check_element_type(e, &seek))
                     {
                         KaxSeekID &seek_id = GetChild<KaxSeekID>(*seek);
-                        EbmlId ebml_id(seek_id.GetBuffer(), (const unsigned int)seek_id.GetSize());
+                        EbmlId ebml_id(seek_id.GetBuffer(), (unsigned int)seek_id.GetSize());
```
xthexder: GCC Warning: `forming offset 16 is out of the bounds [0, 15]`

tests/UnitTests/depthmcu_ut/depthmcu_ut.cpp
```diff
@@ -495,7 +495,7 @@ TEST_F(depthmcu_ut, depthmcu_get_serialnum_non_ascii)
 
     // Create a serial number with non-ACII characters
     char mockSerialNumber[20];
-    memcpy(mockSerialNumber, "12\145\t67890\0WXYZ", sizeof("1234567890\0WXYZ"));
+    memcpy(mockSerialNumber, "12\145\t67890\0WXYZ", sizeof("12\145\t67890\0WXYZ"));
```
xthexder: GCC Warning: `forming offset 16 is out of the bounds [0, 15]`

tools/k4aviewer/k4apointcloudwindow.cpp
```diff
@@ -92,7 +92,7 @@ void K4APointCloudWindow::Show(K4AWindowPlacementInfo placementInfo)
     }
 
     ImGui::SameLine();
-    ImGui::VerticalSeparator();
+    ImGui::SeparatorEx(ImGuiSeparatorFlags_Vertical);
```
williampMSFT: I don't object to this, but I'd like to understand why we're making the change so I can do the right thing in future - what's the deal here?

xthexder: `VerticalSeparator()` was from imgui_internal.h and was changed in the latest release. This is the equivalent behavior in the latest version.

cmake/k4aCompilerFlags.cmake
```diff
@@ -29,6 +29,11 @@ if ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
         # Added in clang 5
         list(APPEND CLANG_ALL_WARNINGS "-Wno-zero-as-null-pointer-constant") # Allow zero as nullptr
     endif()
+    if (NOT (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS "8.0.0"))
```
skalldri: Took me a second to parse what this was trying to accomplish. Consider:

`if (${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL "8.0.0")`

xthexder: It looks like we support CMake 3.5.1 (which comes with Ubuntu 16.04), `VERSION_GREATER_EQUAL` was not added until CMake 3.7, which is why this check was done this way for the above 5.0.0 check as well.

skalldri: Ah... I missed that. Sounds good.

src/record/internal/matroska_read.cpp
schultetwin1: Since this is cpp should we use `static_cast` instead?

```suggestion
                        EbmlId ebml_id(seek_id.GetBuffer(), static_cast<unsigned int>(seek_id.GetSize()));
```

tests/UnitTests/depthmcu_ut/depthmcu_ut.cpp
schultetwin1: Hats off to GCC for catching this!

wes-b:
@xthexder 
> Updated ImGui to v1.72b to fix some warnings

How was this changed? I don't see a submodule commit update?

tests/UnitTests/depthmcu_ut/depthmcu_ut.cpp
wes-b: What is going on with formatting in this string?

PR #666, opened 2019-08-21
Reset now succeeds if Color camera is hung

### LINKED ISSUE #665: AzureKinectFirmwareTool fails to reset device if color camera is hung
**Describe the bug**
AzureKinectFirmwareTool fails to reset device if color camera is hung

**To Reproduce**
1. In Device manager disable color camera
2. >AzureKinectFirmwareTool -r

**Expected behavior**
This should succeed

 - SDK Version:  most resent commit

wes-b:
- Allows firmware_create to succeed if only the depth MCU is found

### I tested changes on: 
- [x] Windows
- [ ] Linux

### DIFF
--- a/src/firmware/firmware.c
+++ b/src/firmware/firmware.c
@@ -162,6 +162,11 @@ k4a_result_t firmware_create(char *device_serial_number, bool resetting_device, 
         if (K4A_SUCCEEDED(result))
         {
             result = TRACE_CALL(colormcu_create(container_id, &firmware->colormcu));
+            if ((resetting_device) && K4A_FAILED(result))
+            {
+                // We only need 1 USB device to reset
+                result = K4A_RESULT_SUCCEEDED;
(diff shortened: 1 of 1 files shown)


src/firmware/firmware.c
```diff
@@ -162,6 +162,11 @@ k4a_result_t firmware_create(char *device_serial_number, bool resetting_device,
         if (K4A_SUCCEEDED(result))
         {
             result = TRACE_CALL(colormcu_create(container_id, &firmware->colormcu));
+            if ((resetting_device) && K4A_FAILED(result))
```
MsDerekMa: Can we double check that colormcu is not being updated if create fails/it will stay null when the create fails?

wes-b: Yes, it is NULL'd out by colormcu_create in the event of a failure. Good question

Issue #683, opened 2019-08-27
Get better performance with depth images in the WinForms example with C# SDK


JorTurFer:
**Is your feature request related to a problem? Please describe.**
When you are tryng the C# SDK with WinForms the maximum frame rate that you can get with the "SetPixel" implementantion is arround 4-5 fps. 
This makes you feel that the problem is the SDK and realy the problem is the implementation of the example.. With the correct implementation the example can run at 15 or 30 fps without any problem.

**Describe the solution you'd like**
The replace this implementantion to access to the bytes directly using `BitmapData` and `Marshal`

JorTurFer:
#686 Closed this issue

PR #684, opened 2019-08-27
Improved perf of .NET samples (Reduced sample ColorResolution to R720p)

reneschulte:
Reduced ColorResolution of the WPF sample to R720p to get max 30fps performance instead of 10fps. The XAML MainWindow has much less height than 720p anyway, so the higher res color frames are just wasted computing.

-
-
-

### I tested changes on: 
- [x] Windows
- [ ] Linux

### DIFF
--- a/src/csharp/Examples/WPF/MainWindow.xaml.cs
+++ b/src/csharp/Examples/WPF/MainWindow.xaml.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Azure.Kinect.Sensor.Examples.WPFViewer
                 device.StartCameras(new DeviceConfiguration
                 {
                     ColorFormat = ImageFormat.ColorBGRA32,
-                    ColorResolution = ColorResolution.R1440p,
+                    ColorResolution = ColorResolution.R720p,
                     DepthMode = DepthMode.WFOV_2x2Binned,
                     SynchronizedImagesOnly = true,
                     CameraFPS = FPS.FPS30,
--- a/src/csharp/Examples/WinForms/Form1.cs
+++ b/src/csharp/Examples/WinForms/Form1.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.Kinect.Sensor.Examples.WinForms
                 device.StartCameras(new DeviceConfiguration
                 {
                     ColorFormat = ImageFormat.ColorBGRA32,
-                    ColorResolution = ColorResolution.R1080p,
+                    ColorResolution = ColorResolution.R720p,
                     DepthMode = DepthMode.NFOV_2x2Binned,
                     SynchronizedImagesOnly = true,
                 });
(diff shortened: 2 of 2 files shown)


JorTurFer:
The problem of the frame rate is not the resolution of the image.
Actualy, in the WinForms example the performance of the depth image generation is too low and it drops the fps.
See the issue #683

reneschulte:
> 
> 
> The problem of the frame rate is not the resolution of the image.
> Actualy, in the WinForms example the performance of the depth image generation is too low and it drops the fps.
> See the issue #683

Surely the pixel copying should be improved and use the same pixel format so a simple mem block copy can be performed. 

Reducing the resolution is just a quick fix/workaround as the window size is smaller anyway.

JorTurFer:
> > The problem of the frame rate is not the resolution of the image.
> > Actualy, in the WinForms example the performance of the depth image generation is too low and it drops the fps.
> > See the issue #683
> 
> Surely the pixel copying should be improved and use the same pixel format so a simple mem block copy can be performed.
> 
> Reducing the resolution is just a quick fix/workaround as the window size is smaller anyway.

Surely... But the correct fix i think that is fix the depth image generation. For example, in the WinForms example I can run with 30 fps only doing it.
image

If you want save the images for anything (to make a video for example), reduce the resolution may be a problem.

reneschulte:
> 
> 
> > > The problem of the frame rate is not the resolution of the image.
> > > Actualy, in the WinForms example the performance of the depth image generation is too low and it drops the fps.
> > > See the issue #683
> > 
> > 
> > Surely the pixel copying should be improved and use the same pixel format so a simple mem block copy can be performed.
> > Reducing the resolution is just a quick fix/workaround as the window size is smaller anyway.
> 
> Surely... But the correct fix i think that is fix the depth image generation. For example, in the WinForms example I can run with 30 fps only doing it.
> image
> 
> If you want save the images for anything (to make a video for example), reduce the resolution may be a problem.

You are right! It should be nicely fixed with the real API usage. I will update the PR. Thanks for the info! 🙌

JorTurFer:
> > > > The problem of the frame rate is not the resolution of the image.
> > > > Actualy, in the WinForms example the performance of the depth image generation is too low and it drops the fps.
> > > > See the issue #683
> > > 
> > > 
> > > Surely the pixel copying should be improved and use the same pixel format so a simple mem block copy can be performed.
> > > Reducing the resolution is just a quick fix/workaround as the window size is smaller anyway.
> > 
> > 
> > Surely... But the correct fix i think that is fix the depth image generation. For example, in the WinForms example I can run with 30 fps only doing it.
> > image
> > If you want save the images for anything (to make a video for example), reduce the resolution may be a problem.
> 
> You are right! It should be nicely fixed with the real API usage. I will update the PR. Thanks for the info! 🙌

I have made a PR with the changes for winforms #686

reneschulte:
Fixed wrong commit

PR #686, opened 2019-08-27
Performance improvement of Winforms' example

JorTurFer:
## Fixes #
#683
- Change the way to generate the depth image from Bitmap.SetPixel to generate byte array and change it

### I tested changes on: 
- [x] Windows
- [ ] Linux

I tested the changes running the device localy:
image

### DIFF
--- a/src/csharp/Examples/WinForms/Form1.cs
+++ b/src/csharp/Examples/WinForms/Form1.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Diagnostics;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using Microsoft.Azure.Kinect.Sensor.WinForms;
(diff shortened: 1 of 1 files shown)


MsDerekMa:
@JorTurFer Thank you for the pull request. I took a look at it and it is indeed something we had planned to do, thanks for implementing it.
I had a couple comments. You should be able to reduce allocations and only allocate your Bitmap and backing buffer once. In addition, using the color object results in a bunch of function calls. So setting the bytes directly is a decent performance improvement. And then lastly I was seeing issues with the window updating the drawing.
Since there are a couple things, I am going to create a pull request into your branch that you can then review.

MsDerekMa:
Apparently GitHub won't let me create a PR/suggestion for the file. So I just committed it.  Please let me know if these changes look good to you.

JorTurFer:
Hi @MsDerekMa !
These changes feel great. From my side, all it's right with them.

PR #720, opened 2019-09-04
Green screen should timeout if it can not get sync'd images

### LINKED ISSUE #719: Green screen example is failing is automated testing
**Describe the bug**
The test is failing in 2 ways
* getting a synch'd sample is running forever 
* on Windows the test is failing to return a clean exit code.

**Expected behavior**
It should pass tests

wes-b:
- Added a timeout of 60s in the event the example can not sync images
- Added debug output to furthar debug the problem
- Removed all 'throw' and replaced with exit(1) (for error)
- Added return 0 to the end of main when the app runs to completion.

### I tested changes on: 
- [ ] Windows
- [ ] Linux

### DIFF
--- a/examples/green_screen/MultiDeviceCapturer.h
+++ b/examples/green_screen/MultiDeviceCapturer.h
@@ -14,7 +14,7 @@ constexpr std::chrono::microseconds MAX_ALLOWABLE_TIME_OFFSET_ERROR_FOR_IMAGE_TI
 
 constexpr int64_t WAIT_FOR_SYNCHRONIZED_CAPTURE_TIMEOUT = 60000;
 
-static void log_lagging_time(char *lagger, k4a::capture &master, k4a::capture &sub)
+static void log_lagging_time(const char *lagger, k4a::capture &master, k4a::capture &sub)
 {
     std::cout << std::setw(6) << lagger << " lagging: mc:" << std::setw(6)
               << master.get_color_image().get_device_timestamp().count() << "us sc:" << std::setw(6)
--- a/examples/green_screen/MultiDeviceCapturer.h
+++ b/examples/green_screen/MultiDeviceCapturer.h
@@ -14,14 +14,14 @@ constexpr std::chrono::microseconds MAX_ALLOWABLE_TIME_OFFSET_ERROR_FOR_IMAGE_TI
 
 constexpr int64_t WAIT_FOR_SYNCHRONIZED_CAPTURE_TIMEOUT = 60000;
 
-void log_lagging_time(char *lagger, k4a::capture &master, k4a::capture &sub)
+static void log_lagging_time(char *lagger, k4a::capture &master, k4a::capture &sub)
 {
     std::cout << std::setw(6) << lagger << " lagging: mc:" << std::setw(6)
               << master.get_color_image().get_device_timestamp().count() << "us sc:" << std::setw(6)
               << sub.get_color_image().get_device_timestamp().count() << "us\n";
 }
 
-void log_synced_image_time(k4a::capture &master, k4a::capture &sub)
+static void log_synced_image_time(k4a::capture &master, k4a::capture &sub)
 {
     std::cout << "Sync'd capture: mc:" << std::setw(6) << master.get_color_image().get_device_timestamp().count()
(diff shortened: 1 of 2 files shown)


examples/green_screen/MultiDeviceCapturer.h
```diff
@@ -4,13 +4,26 @@
...
 #include <k4a/k4a.hpp>
 
 // This is the maximum difference between when we expected an image's timestamp to be and when it actually occurred.
 // TODO waiting on a firmware update to be returned to 50
 constexpr std::chrono::microseconds MAX_ALLOWABLE_TIME_OFFSET_ERROR_FOR_IMAGE_TIMESTAMP(33000);
 // constexpr std::chrono::microseconds MAX_ALLOWABLE_TIME_OFFSET_ERROR_FOR_IMAGE_TIMESTAMP(50);
 
+constexpr int64_t WAIT_FOR_SYNCHRONIZED_CAPTURE_TIMEOUT = 60000;
+
+#define LOG_LAGGING_TIME(lagger)                                                                                       \
```
schultetwin1: Just a thought, is there any benefit to using a macro here over a function? We could make `LOG_LAGGING_TIME` and `LOG_SYNCED_IMAGE_TIME` static functions instead of macros. This gives us better type checking, more readable compile errors, and easier stack tracking?

examples/green_screen/MultiDeviceCapturer.h
```diff
@@ -20,7 +33,8 @@ class MultiDeviceCapturer
         bool master_found = false;
         if (device_indices.size() == 0)
         {
-            throw std::runtime_error("Capturer must be passed at least one camera!");
+            cerr << "Capturer must be passed at least one camera!\n ";
```
schultetwin1: For my own learning, why remove the throws and instead just `exit(1)`?

examples/green_screen/MultiDeviceCapturer.h
```diff
@@ -4,13 +4,26 @@
...
 #include <vector>
+#include <iomanip> // std::setw
 #include <k4a/k4a.hpp>
 
 // This is the maximum difference between when we expected an image's timestamp to be and when it actually occurred.
 // TODO waiting on a firmware update to be returned to 50
 constexpr std::chrono::microseconds MAX_ALLOWABLE_TIME_OFFSET_ERROR_FOR_IMAGE_TIMESTAMP(33000);
 // constexpr std::chrono::microseconds MAX_ALLOWABLE_TIME_OFFSET_ERROR_FOR_IMAGE_TIMESTAMP(50);
 
+constexpr int64_t WAIT_FOR_SYNCHRONIZED_CAPTURE_TIMEOUT = 60000;
```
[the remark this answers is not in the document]
wes-b: It tells the compiler it can be evaluated at compile time like a #define

examples/green_screen/MultiDeviceCapturer.h
wes-b: For testing, main should return 0 for success. I was not sure if throws were randomly setting the return code so I set them explicitly. I was also not seeing the error message in the console output when throw was firing so the associated message was not helping.

examples/green_screen/MultiDeviceCapturer.h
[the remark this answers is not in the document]
schultetwin1: Ah, `const` does not guarantee compile time evaluation and `constexpr` does. Super interesting, thanks!

examples/green_screen/MultiDeviceCapturer.h
schultetwin1: To add to the learning:

According to [cppreference](https://en.cppreference.com/w/cpp/error/terminate) when an exception is thrown and not caught, `std::terminate` is called. The default implementation of `std::terminate` is to call `std::abort` (though the terminate handler can be overridden at run time by the program). According to [cppreference](https://en.cppreference.com/w/cpp/utility/program/abort), `std::abort` causes abnormal program termination. Therefore, it should not set the exit code to 0. In fact, on Windows, `std::abort` results in an exit code of 3. I was unable to determine the exit code on Linux.

I'm fine with doing error handling this way. If we wanted to use the stack unwinding, we could also just catch `std::runtime_error` where ever the constructor for `MultiDeviceCapture` is.

PR #730, opened 2019-09-05
Dynamically link the C++ runtime

### LINKED ISSUE #728: Green Screen crashes when debug build is running
**Describe the bug**
This is due to building a debug k4a.dll and linking to a retail version of open cv

### LINKED ISSUE #737: Switch to dynamically link against the C++ runtime
Currently, k4a.dll is statically linked against the C++ run-time library. This was originally done in order to remove the need of shipping the proper run-time library with k4a.dll. However, this has a couple downsides:

* If users are using other DLLs that link against the dynamic C++ runtime, we may run into issues with data structures passed between those DLLs and k4a.dll. A specific example is if another DLL allocates memory which we try to free in k4a.dll.

* It bloats the size of our DLL (which maybe unnecessary if the users of our library already link dynamically against the C++ runtime).

* It provides a security risk for us if a security issue in the C++ runtime is found. 

Because of these issues we would like to change k4a.dll to dynamically link against the C++ runtime.

**Additional context**
As part of this change, we should also link the depthengine dll against the dynamic version of the visual C++ runtime. @cdedmonds and @yijiew as a FYI.

wes-b:
- Changed linking for debug builds for Green Screen and Open CV Compatibility
- Changed linking for all all projects to not static link VC binaries.


wes-b:
With this change I am currently getting the following error:
```
LINK : warning LNK4098: defaultlib 'LIBCMTD' conflicts with use of other libs; use /NODEFAULTLIB:library
```
 for:
* transformation_example.exe
* test_dynlib_1_2.dll
* k4arecorder.exe
and others

PR #731, opened 2019-09-05
OpenCV does not need to be in users PATH

### LINKED ISSUE #703: Copy the OpenCV dlls into the relevant directory so that users don't have to change their PATH

Before I left, we discussed the possibility of avoiding the update to the PATH by simply copying the relevant OpenCV DLLs into the bin directory.

If we wanted to do this, it would require:
- Editing our CMake setup so that it copies over those files based on OpenCV_DIR
- Changing azure-pipelines.yml so that the step that updates the PATH is removed
- Changing the README for green_screen which says to add it to your PATH

**Describe alternatives you've considered**

We could also just remain using the PATH option. It is, in fairness, a reasonable way to do this.

**Additional context**

OpenCV's suggested installation method modifies your PATH

wes-b:
- Copy DLL's to bin folder so path updates are not required.
- Updated readme.md to reflect changes.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -22,6 +22,6 @@ if (OpenCV_FOUND)
     add_subdirectory(opencv_compatibility)
 
     if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
-        file(COPY ${OpenCV_DIR}/../bin DESTINATION ${CMAKE_BINARY_DIR} FILES_MATCHING PATTERN "*.dll")
+        file(COPY ${OpenCV_DIR}/../bin/ DESTINATION ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} FILES_MATCHING PATTERN "*.dll")
     endif ()
 endif ()
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -20,4 +20,8 @@ endif ()
 if (OpenCV_FOUND)
     add_subdirectory(green_screen)
     add_subdirectory(opencv_compatibility)
+
+    if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+        file(COPY ${OpenCV_DIR}/../bin DESTINATION ${CMAKE_BINARY_DIR} FILES_MATCHING PATTERN "*.dll")
+    endif ()
 endif ()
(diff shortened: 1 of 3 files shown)


examples/CMakeLists.txt
```diff
@@ -20,4 +20,8 @@ endif ()
 if (OpenCV_FOUND)
     add_subdirectory(green_screen)
     add_subdirectory(opencv_compatibility)
+
+    if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+        file(COPY ${OpenCV_DIR}/../bin DESTINATION ${CMAKE_BINARY_DIR} FILES_MATCHING PATTERN "*.dll")
```
schultetwin1: I think we actually want to use `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}`
```suggestion
        file(COPY ${OpenCV_DIR}/../bin DESTINATION ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} FILES_MATCHING PATTERN "*.dll")
```

`${CMAKE_BINARY_DIR}` is just the /build/ folder, `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}` is the /build/bin folder

examples/green_screen/README.md
```diff
@@ -45,11 +46,9 @@ This example requires OpenCV to be installed to build. To ensure it will be buil
 #### Windows
 
 Our recommended way of getting OpenCV on Windows is by installing pre-built libraries. There's even a PowerShell script
-that'll do much of the work for you in `scripts/install-opencv.ps1`. This will place OpenCV in your `C:\` folder. You'll
-also need to add the folder where the OpenCV DLLs are to your PATH. If you used the script, that means you need to add
-`C:\opencv\build\x64\vc14\bin` to your PATH. If you installed OpenCV somewhere else, you may need to add the
-`-DOpenCV_DIR=path\to\a\folder\in\your\opencv\install\that\contains\the\file\OpenCVConfig.cmake` option to cmake when
-you build.
+that'll do much of the work for you in `scripts/install-opencv.ps1`. This will place OpenCV in your `C:\` folder. CMake
+will copy the OpenCV binaries to the the `${CMAKE_BINARY_DIR}` folder so that all built binaries will run as expected. A
```
schultetwin1:
```suggestion
will copy the OpenCV binaries to the the `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}` folder so that all built binaries will run as expected. A 
```

PR #749, opened 2019-09-10
Add a device::handle() function to k4a.hpp

### LINKED ISSUE #748: k4a.hpp cannot access the k4a_device_t handle from the device class

I want to use the C++ wrapper in recording API.
<code>k4a_record_create()</code> requests a <code>k4a_device_t</code> handle in second argument.
However, There is no way we can access the <code>k4a_device_t</code> handle from the <code>k4a::device</code> class.
That means we can't use recording API with C++ wrapper.

Create a <cpde>handle()</code> function like the image class that expose allow users to access the k4a_device_t handle from the <code>k4a::device</code> class.

```patch
+ k4a_device_t handle() const noexcept
+ {
+     return m_handle;
+ }
```

UnaNancyOwen:
- Adds a device::handle() method to the C++ wrapper, allowing access to the underlying k4a_device_t when using the C++ wrapper. This is needed to allow the C++ wrapper and the recording API to work together.

  Specifically, <code>k4a_record_create()</code> requests a <code>k4a_device_t</code> handle in second argument.
```cpp
k4a_record_t recording_handle = nullptr;
k4a_record_create( path, device.handle(), device_configuration, &recording_handle );
```

### I tested changes on: 
- [x] Windows
- [ ] Linux

### DIFF
--- a/include/k4a/k4a.hpp
+++ b/include/k4a/k4a.hpp
@@ -1116,6 +1116,18 @@ public:
         return m_handle != nullptr;
     }
 
+    /** Returns the underlying k4a_device_t handle
+     *
+     * Note that this function does not increment the reference count on the k4a_device_t.
+     * The caller is responsible for incrementing the reference count on
+     * the k4a_device_t if the caller needs the k4a_device_t to outlive this C++ object.
+     * Otherwise, the k4a_device_t will be destroyed by this C++ object.
+     */
+    k4a_device_t handle() const noexcept
+    {
+        return m_handle;
+    }
+
     /** Closes a k4a device.
      *
      * \sa k4a_device_close
--- a/include/k4a/k4a.hpp
+++ b/include/k4a/k4a.hpp
@@ -1118,10 +1118,8 @@ public:
 
     /** Returns the underlying k4a_device_t handle
      *
-     * Note that this function does not increment the reference count on the k4a_device_t.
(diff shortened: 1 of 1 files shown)


include/k4a/k4a.hpp
```diff
@@ -1116,6 +1116,18 @@ class device
         return m_handle != nullptr;
     }
 
+    /** Returns the underlying k4a_device_t handle
+     *
+     * Note that this function does not increment the reference count on the k4a_device_t.
+     * The caller is responsible for incrementing the reference count on
+     * the k4a_device_t if the caller needs the k4a_device_t to outlive this C++ object.
+     * Otherwise, the k4a_device_t will be destroyed by this C++ object.
```
wes-b: Thank you for submitting this. The handle is not reference counted at all and there is unfortunately no way to  make sure the handle survives destruction. I would say something like: _** Note the k4a_device_t handle does not have a reference count will be destroyed when the C++ object is destroyed. The caller is responsible for ensuring the C++ object outlives this handle.**_

wes-b:
@UnaNancyOwen you good with this change?

include/k4a/k4a.hpp
UnaNancyOwen: Thanks! Yes, that's right!

UnaNancyOwen:
@wes-b Good! 👍

Issue #764, opened 2019-09-12
Depth Engine appears to be leaking resource to d3d


wes-b:
**Describe the bug**
On process exit if the depth engine has been running, an exception will be thrown, durring rundown, as a result of d3d11!UnregisterLeakedDeviceTrimNotifications being called. If I run without depth engine running I get a clean exit CPP DTOR's run, and we get leak analysis.

**To Reproduce**
1. Run the gtest below and inspect the exit callstack in windbg

**gtest**
```
TEST_F(multidevice_ft, open_parallel)
{

    ASSERT_EQ((uint32_t)2, k4a_device_get_installed_count());

    k4a_device_t device;
    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_open(0, &device));

    k4a_device_configuration_t config = K4A_DEVICE_CONFIG_INIT_DISABLE_ALL;
    config.color_format = K4A_IMAGE_FORMAT_COLOR_MJPG;
    config.color_resolution = K4A_COLOR_RESOLUTION_OFF;
    config.depth_mode = K4A_DEPTH_MODE_NFOV_2X2BINNED;
    config.camera_fps = K4A_FRAMES_PER_SECOND_30;

    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_start_cameras(device, &config));
    // ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_start_imu(device));

    k4a_device_close(device);
}
```
**Expected behavior**
```
00 00000097`364ffd08 00007ff9`3fbcdcd8 ntdll!ZwTerminateProcess
01 00000097`364ffd10 00007ff9`3f14e0ab ntdll!RtlExitUserProcess
02 00000097`364ffd40 00007ff8`cf5e3e4a KERNEL32!FatalExit
03 00000097`364ffd70 00007ff8`cf5e3df9 ucrtbased!exit_or_terminate_process
04 00000097`364ffda0 00007ff8`cf5e4176 ucrtbased!common_exit
05 00000097`364ffe00 00007ff7`2c521867 ucrtbased!exit
06 00000097`364ffe30 00007ff7`2c52170e multidevice_ft!__scrt_common_main_seh
07 00000097`364ffea0 00007ff7`2c521999 multidevice_ft!__scrt_common_main
08 00000097`364ffed0 00007ff9`3f147034 multidevice_ft!mainCRTStartup
[...]

PR #766, opened 2019-09-13
Depth loader thread safety

### LINKED ISSUE #676: Thread safety for k4a_device_start_cameras()
Camera starting is not thread safe (cannot open/start different cameras in parallel).

Mainly seems because the DLL dynamic loader does not have a mutex around this context object:
deloader_context_t g_deloader

wes-b:
- Add global context to ensure global is only init'd once
- Added static CPP class and DTOR to handle cleanup (#764 prevents this from running 100% of the time).
- Added functional tests to confirm fix

### I tested changes on: 
- [x] Windows
- [ ] Linux


tests/multidevice/multidevice.cpp
```diff
@@ -224,4 +224,155 @@ TEST_F(multidevice_ft, ensure_color_camera_is_enabled)
...
+    bool started;
+    LOCK_HANDLE lock;
+} parallel_start_data_t;
+
+static int parallel_start_thread(void *param)
+{
+    parallel_start_data_t *data = (parallel_start_data_t *)param;
+    k4a_result_t result = K4A_RESULT_SUCCEEDED;
+
+    Lock(data->lock);
```
MsDerekMa: Because you are using this for synchronization, one thread will always be slightly ahead of the other by however long the acquire and unlock of lock is. What this close enough to cause the old implementation to fail?

tests/multidevice/multidevice.cpp
```diff
@@ -224,4 +224,155 @@ TEST_F(multidevice_ft, ensure_color_camera_is_enabled)
...
+
+    data2.config = data1.config = &config;
+    data1.lock = data2.lock = lock;
+
+    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_start_cameras(data1.device, data1.config));
+    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_start_cameras(data2.device, data2.config));
+    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_start_imu(data1.device));
+    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_start_imu(data2.device));
+
+    data1.started = data2.started = true;
```
MsDerekMa: Were parallel closes causing an issue before? It seems like we weren't closing the dll handle before and the engine destroy flow shouldn't really be affected.

tests/multidevice/multidevice.cpp
```diff
@@ -224,4 +224,155 @@ TEST_F(multidevice_ft, ensure_color_camera_is_enabled)
...
+TEST_F(multidevice_ft, start_parallel)
+{
+    LOCK_HANDLE lock;
+    THREAD_HANDLE th1, th2;
+    ASSERT_NE((lock = Lock_Init()), (LOCK_HANDLE)NULL);
+    parallel_start_data_t data1 = { 0 }, data2 = { 0 };
+
+    ASSERT_EQ((uint32_t)2, k4a_device_get_installed_count());
+
+    ASSERT_EQ(K4A_RESULT_SUCCEEDED, k4a_device_open(0, &data1.device));
```
MsDerekMa: Why aren't you doing the open inside the parallel threads?

tests/multidevice/multidevice.cpp
wes-b: Yes it was, I was able repro it very easily.

tests/multidevice/multidevice.cpp
wes-b: No, parallel closes were not an issue before. And to your point we were not fully closing the handle so closes were unknown. This test establishes that we don't have issues with parallel closes.

tests/multidevice/multidevice.cpp
wes-b: Parallel open is currently not supported. 2 separate LibUsb handles are required and that would need to be serialized. I don't see a great way to implement that yet, other than just creating another global lock.  The solution has to take into account the k4a_device_open, k4a_device_get_installed_count, and AzureKinectFirmwareTool where they each open the two handles in slightly different ways.

PR #785, opened 2019-09-23
Add C++ wrapper for recording API

### LINKED ISSUE #751: Add C++ wrapper for Recording API
Currently, There is no C++ wrapper for Recording API yet.
Do you have any plan to add record.hpp? Thanks,

In relation to this, I tried to pre-implement C++ wrapper for recording API.
It is need k4a::device::handle(). Please read this issue https://github.com/microsoft/Azure-Kinect-Sensor-SDK/issues/748.
https://gist.github.com/UnaNancyOwen/4addf99a1a712195ac399c28bfa334ef#file-record-hpp

UnaNancyOwen:
- Add C++ wrapper for recording API.
  Simple usage is as follows.
  
```cpp
// Include Header
#include <k4a/k4a.hpp>
#include <k4arecord/record.hpp>

// Open Device and Start Cameras
k4a::device device = k4a::device::open(device_index);
k4a_device_configuration_t device_configuration = K4A_DEVICE_CONFIG_INIT_DISABLE_ALL;
device_configuration.color_format             = K4A_IMAGE_FORMAT_COLOR_BGRA32;
device_configuration.color_resolution         = K4A_COLOR_RESOLUTION_720P;
device_configuration.depth_mode               = K4A_DEPTH_MODE_NFOV_UNBINNED;
device_configuration.synchronized_images_only = true;
device.start_cameras(&device_configuration);

// Create Record
k4a::record record = k4a::record::create("record.mkv", device, device_configuration);

// Write Header
record.write_header();

while(is_record)
{
    // Get Capture
    k4a::capture capture;
    device.get_capture(&capture, K4A_WAIT_INFINITE);
    
    // Write Capture
    record.write_capture(capture);

    // Release Handle
    capture.reset();
}

// Flush and Close Record
record.flush();
record.close();

// Stop Cameras and Close Device
device.stop_cameras();
device.close();
```

### I tested changes on: 
- [x] Windows
- [ ] Linux

### DIFF
--- a/include/k4arecord/record.hpp
+++ b/include/k4arecord/record.hpp
@@ -40,6 +40,7 @@ public:
 
     ~record()
     {
+        // NOTE: flush is called internally when closing record
         close();
     }
 
--- a/include/k4arecord/record.hpp
+++ b/include/k4arecord/record.hpp
@@ -252,7 +252,7 @@ public:
 
         if (K4A_RESULT_SUCCEEDED != result)
         {
-            throw error("Failed to write custom_track_data!");
+            throw error("Failed to write custom track data!");
         }
     }
 
--- a/include/k4arecord/record.hpp
+++ b/include/k4arecord/record.hpp
@@ -40,7 +40,6 @@ public:
 
     ~record()
     {
-        flush();
         close();
     }
 
@@ -52,7 +51,6 @@ public:
     {
         if (this != &other)
         {
-            flush();
             close();
             m_handle = other.m_handle;
             other.m_handle = nullptr;
@@ -89,7 +87,12 @@ public:
     {
         if (m_handle)
         {
-            k4a_record_flush(m_handle);
+            k4a_result_t result = k4a_record_flush(m_handle);
+
+            if (K4A_RESULT_SUCCEEDED != result)
+            {
+                throw error("Failed to flush!");
+            }
         }
     }
 
--- a/include/k4arecord/record.hpp
+++ b/include/k4arecord/record.hpp
@@ -249,7 +249,7 @@ public:
 
         if (K4A_RESULT_SUCCEEDED != result)
         {
-            throw error("Failed to write imu sample!");
+            throw error("Failed to write custom_track_data!");
         }
     }
 
--- a/include/k4arecord/record.hpp
+++ b/include/k4arecord/record.hpp
@@ -1,0 +1,280 @@
+/** \file record.hpp
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ * Kinect For Azure SDK - C++ wrapper.
+ */
+#ifndef K4A_RECORD_HPP
+#define K4A_RECORD_HPP
+
+#include <k4a/k4a.hpp>
+#include <k4arecord/record.h>
+
+namespace k4a
+{
+
+/** \class record record.hpp
+ * Wrapper for \ref k4a_record_t
+ *
+ * Wraps a handle for a record object
+ *
+ * \sa k4a_record_t
+ */
+class record
+{
+public:
+    /** Creates a k4a::record from a k4a_record_t
+     * Takes ownership of the handle, i.e. you should not call
+     * k4a_record_close on the handle after giving it to the
+     * k4a::record; the k4a::record will take care of that.
+     */
+    record(k4a_record_t handle = nullptr) noexcept : m_handle(handle) {}
+
+    /** Moves another k4a::record into a new k4a::record
+     */
+    record(record &&other) noexcept : m_handle(other.m_handle)
+    {
+        other.m_handle = nullptr;
+    }
+
+    record(const record &) = delete;
+
+    ~record()
+    {
+        flush();
+        close();
+    }
+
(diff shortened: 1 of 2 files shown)


include/k4arecord/record.hpp
```diff
@@ -0,0 +1,280 @@
...
+            throw error("Failed to write imu sample!");
+        }
+    }
+
+    /** Opens a new recording file for writing
+     * Throws error on failure
+     *
+     * \sa k4a_record_create
+     */
+    static record create(const char *path, device &device, k4a_device_configuration_t &device_configuration)
```
UnaNancyOwen: I wondered about why k4a::device don't hold k4a_device_configuration_t?
If it is holded, this function doesn't need to input k4a_device_configuration_t.
(Or, I think it is good add function to get k4a_device_configuration_t from k4a_device_t.)

wes-b: Storing the configuration would allow the create API for record to eliminate one parameter, but the SDK would have to keep track if k4a_device_start_cameras has been called to provide the config. If the user calls record::create before start camera's then should it it just fail.

Another issue we would have to be concerned with is after ::stop_cameras() is called and the configuration is known by the SDK. Simple enough to null out, but we would need to test and ensure that works as expected. 

I think for now it would simplier to just pass in the config and not create any dependencies between the two API's and just pass the parameter into each API.

include/k4arecord/record.hpp
```diff
@@ -0,0 +1,280 @@
...
+    record(record &&other) noexcept : m_handle(other.m_handle)
+    {
+        other.m_handle = nullptr;
+    }
+
+    record(const record &) = delete;
+
+    ~record()
+    {
+        flush();
```
wes-b: Think we should just move flush into the close function. I don't see a scenario where you don't want that called.

If we don't, then call to flush should be wrapped in a `if (m_handle){}` check

@xthexder, what do you think?

include/k4arecord/record.hpp
```diff
@@ -0,0 +1,280 @@
...
+
+    /** Flushes all pending recording data to disk
+     *
+     * \sa k4a_record_flush
+     */
+    void flush()
+    {
+        if (m_handle)
+        {
+            k4a_record_flush(m_handle);
```
wes-b: This function can fail, so we need a throw here if this fails.

include/k4arecord/record.hpp
```diff
@@ -0,0 +1,280 @@
...
+    {
+        k4a_result_t result = k4a_record_write_custom_track_data(m_handle,
+                                                                 track_name,
+                                                                 device_timestamp_usec,
+                                                                 custom_data,
+                                                                 custom_data_size);
+
+        if (K4A_RESULT_SUCCEEDED != result)
+        {
+            throw error("Failed to write imu sample!");
```
wes-b:
```suggestion
            throw error("Failed to write custom_track_data!");
```

include/k4arecord/record.hpp
xthexder: I agree with Wes on this. The current API also allows for recordings to be generated without a device present, which is useful for synthetic data testing.

include/k4arecord/record.hpp
xthexder: k4a_record_close() will already call flush internally, but since it's a void function, errors are ignored.
Flush was made accessible as a separate function so the user can flush periodically and verify data was written to disk.
If we're not handling flush errors, then there is no point to calling flush separately.

UnaNancyOwen: I see. I remove flush from this destructor.

include/k4arecord/record.hpp
UnaNancyOwen: It is my mistake. Add throw an error.

include/k4arecord/record.hpp
UnaNancyOwen: I got it. Thanks,

include/k4arecord/record.hpp
```diff
@@ -0,0 +1,283 @@
...
+    {
+        k4a_result_t result = k4a_record_write_custom_track_data(m_handle,
+                                                                 track_name,
+                                                                 device_timestamp_usec,
+                                                                 custom_data,
+                                                                 custom_data_size);
+
+        if (K4A_RESULT_SUCCEEDED != result)
+        {
+            throw error("Failed to write custom_track_data!");
```
wes-b:
```suggestion
            throw error("Failed to write custom track data!");
```

include/k4arecord/record.hpp
```diff
@@ -0,0 +1,280 @@
...
+    {
+        other.m_handle = nullptr;
+    }
+
+    record(const record &) = delete;
+
+    ~record()
+    {
+        flush();
+        close();
```
wes-b: Please add a comment that says flush is called on close so that we document what we learned.

UnaNancyOwen: Is this comment correct?
```cpp
~record()
{
    // flush is called on closing record
    close();
}
```

wes-b: Yes :)

PR #799, opened 2019-09-25
Restrict MSSE flags to arch that support it

### LINKED ISSUE #798: `-msse4.1` flag being applied to all Linux builds
**Describe the bug**
The following code snippet

https://github.com/microsoft/Azure-Kinect-Sensor-SDK/blob/2a8a5a324f9befdc58a406fb9dd08a88841a515d/src/transformation/CMakeLists.txt#L19-L21

Applies the `-msse4.1` flag to all Clang or GNU compilers. If should apply to all Clang or GNU compilers that are targeting x86_64 or i386 architectures.

**To Reproduce**
1. Clone our code on a arm machine
2. Build for arm
3. Hit a build error about flag `-msse4.1`

**Expected behavior**
Architectures which don't support msse4.1 should not use msse4.1

schultetwin1:
- Restricts `-msse4.1` flags to processors which (might) support it

### I tested changes on:
- [ ] Windows
- [ ] Linux

I built this change under Linux and made sure the `-msse4.1` flag still existed.

### DIFF
--- a/src/transformation/CMakeLists.txt
+++ b/src/transformation/CMakeLists.txt
@@ -17,7 +17,9 @@ target_link_libraries(k4a_transformation PUBLIC
     )
 
 if ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
-    target_compile_options(k4a_transformation PRIVATE "-msse4.1")
+    if ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "amd64.*|x86_64.*|AMD64.*|i686.*|i386.*|x86.*|amd64.*|AMD64.*")
+        target_compile_options(k4a_transformation PRIVATE "-msse4.1")
+    endif()
 endif()
 
 # Define alias for other targets to link against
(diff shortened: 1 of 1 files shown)


src/transformation/CMakeLists.txt
```diff
@@ -17,7 +17,9 @@ target_link_libraries(k4a_transformation PUBLIC
     )
 
 if ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
-    target_compile_options(k4a_transformation PRIVATE "-msse4.1")
+    if ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "amd64.*|x86_64.*|AMD64.*|i686.*|i386.*|x86.*|amd64.*|AMD64.*")
```
wes-b: Does case matter here? `amd64.*` is listed twice, so is `AMD64.*`

skalldri: At a minimum here, it does look like `amd64` and `AMD64` are duplicated.

schultetwin1: Case does matter here but good call on the `amd64`, fixed that.

PR #802, opened 2019-09-26
Changes made to emphasize Depth Engine version and deemphasize plugin…

### LINKED ISSUE #801: SDK is not reporting the depth engine version correctly.
It is report the plugin version instead of the DE version.

wes-b:
… version. Added trace message to Windows and Linux to log path to loaded depth engine.

- Depth Engine now reports the depth engine version rather than the plugin version.
- Changes made to emphasize Depth Engine version and deemphasize plugin version.
- Added trace message to Windows and Linux to log path to loaded depth engine.
- Also synchronized changes in k4aplugin.h between the two repro's.

### DIFF
--- a/include/k4ainternal/dynlib.h
+++ b/include/k4ainternal/dynlib.h
@@ -42,8 +42,7 @@ K4A_DECLARE_HANDLE(dynlib_t);
  * The version information should be encoded in the filename of the dynamic
  * library being loaded. This encoding will be different for Windows versus Linux. For
  * Windows the dynamic library name is "<name>_<version>_0.dll".
- * For Linux, the dynamic library name is
- * "lib<name>.so.<version>".
+ * For Linux, the dynamic library name is "lib<name>.so.<version>.0".
  *
  * \remarks
  * This function only supports relative path libraries. If you try to load an
--- a/include/k4ainternal/dynlib.h
+++ b/include/k4ainternal/dynlib.h
@@ -20,14 +20,9 @@ extern "C" {
 K4A_DECLARE_HANDLE(dynlib_t);
 
 /**
- * The maximum major version we support for loading a dynamic library
+ * The maximum version we support for loading a dynamic library
  */
-#define DYNLIB_MAX_MAJOR_VERSION 99ul
-
-/**
- * The maximum minor version we support for loading a dynamic library
- */
-#define DYNLIB_MAX_MINOR_VERSION 99ul
+#define DYNLIB_MAX_VERSION 99ul
 
 /** Loads a versioned dynamic library (shared library) by name and version.
  *  The version information is encoded in the filename.
@@ -35,13 +30,9 @@ K4A_DECLARE_HANDLE(dynlib_t);
  * \param name [IN]
  * Name of the dynamic library or shared library
  *
- * \param major_ver
- * Expected major version of the dynamic library. This number must be less than
- * or equal to \ref DYNLIB_MAX_MAJOR_VERSION.
- *
- * \param minor_ver
- * Expected minor version of the dynamic library. This number must be less than
- * or equal to \ref DYNLIB_MIN_MAJOR_VERSION.
+ * \param version
+ * Version of the plugin being used to load the dynamic library. This number must be less than
+ * or equal to \ref DYNLIB_MAX_VERSION.
  *
  * \param dynlib_handle [OUT]
  * A handle to store dynlib in. Only valid if function returns
@@ -49,10 +40,10 @@ K4A_DECLARE_HANDLE(dynlib_t);
  *
  * \remarks
  * The version information should be encoded in the filename of the dynamic
- * library. This encoding will be different for Windows versus Linux. For
- * Windows the dynamic library name is "<name>_<major_ver>_<minor_ver>.dll".
+ * library being loaded. This encoding will be different for Windows versus Linux. For
+ * Windows the dynamic library name is "<name>_<version>_0.dll".
  * For Linux, the dynamic library name is
- * "lib<name>.so.<major_ver>.<minor_ver>".
+ * "lib<name>.so.<version>".
  *
  * \remarks
  * This function only supports relative path libraries. If you try to load an
@@ -64,7 +55,7 @@ K4A_DECLARE_HANDLE(dynlib_t);
  * \return K4A_RESULT_SUCCEEDED if the library was loaded, K4A_RESULT_FAILED
  * otherwise
  */
-k4a_result_t dynlib_create(const char *name, uint32_t major_ver, uint32_t minor_ver, dynlib_t *dynlib_handle);
+k4a_result_t dynlib_create(const char *name, uint32_t version, dynlib_t *dynlib_handle);
 
 /** Finds the address of an exported symbol in a loaded dynamic library
  *
--- a/include/k4ainternal/k4aplugin.h
+++ b/include/k4ainternal/k4aplugin.h
@@ -23,10 +23,10 @@ extern "C" {
 #endif
 
 /**
- * Current Version of the Azure Kinect Depth Engine Plugin
+ * Current Version of the Azure Kinect Depth Engine Plugin Interface
  *
  * \remarks
- * When the depth engine plugin is updated this version should be increased.
+ * When the depth engine plugin interface (k4a_plugin_t) is updated, this version should be increased.
  *
  * \remarks
  * The depth version binary name has the plugin version appended to it to signify the compatibility between the plugin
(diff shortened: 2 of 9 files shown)


include/k4ainternal/k4aplugin.h
```diff
@@ -23,13 +23,28 @@ extern "C" {
...
 
 /**
- * Current Version of the k4A Depth Engine Plugin API
+ * Current Version of the K4A depth engine and this Plugin interface
  *
- * \remarks The MAJOR version must be updated to denote any breaking change
- * that would cause an older SDK to not be able to use this plugin.
+ * \remarks
+ * This plugin version and the depth engine are tied together with the major.minor version. When the depth engine takes
+ * a breaking change, then its major version must me increased. In turn the major version of this interface must
```
schultetwin1:
```suggestion
 * a breaking change, then its major version must be increased. In turn the major version of this interface must
```

include/k4ainternal/k4aplugin.h
```diff
@@ -23,13 +23,28 @@ extern "C" {
 #endif
 
 /**
- * Current Version of the k4A Depth Engine Plugin API
+ * Current Version of the K4A depth engine and this Plugin interface
  *
- * \remarks The MAJOR version must be updated to denote any breaking change
- * that would cause an older SDK to not be able to use this plugin.
+ * \remarks
+ * This plugin version and the depth engine are tied together with the major.minor version. When the depth engine takes
```
schultetwin1: These remarks mark a difference in the way I was thinking about the depth engine. I think I see the way you are looking at it, but I just want to make sure we are on the same page.

My view
The K4A SDK defines a plugin interface. Anyone can build an implementation against that interface. Since the SDK defines the interface, the SDK also defines the interface version. Any implementation of the interface should take its major and minor version from the interface version defined by the SDK.

What I believe I reading here
The depth engine is the golden standard for the plugin interface. We build an interface to allow others to implement there own plugin, but the root of truth is our depth engine. If our depth engine version changes, so does our plugin interface.

I'm fine either way, just want to make sure I'm understanding correctly.

MsDerekMa:
>  * depthengine_2_0.dll on Windows and libdepthengine.so.2.0 on Linux

Are we keeping the .0 for the Linux binaries?

---
Refers to: include/k4ainternal/dynlib.h:53 in 5c8b541. [](commit_id = 5c8b54178cc4066592f5fb2c71364be7480756b6, deletion_comment = False)

src/dynlib/dynlib_linux.c
```diff
@@ -21,11 +22,11 @@ typedef struct _dynlib_context_t
...
 
 K4A_DECLARE_CONTEXT(dynlib_t, dynlib_context_t);
-static char *generate_file_name(const char *name, uint32_t major_ver, uint32_t minor_ver)
+static char *generate_file_name(const char *name, uint32_t version)
 {
     const char *lib_prefix = "lib";
     const char *lib_suffix = "so";
-    size_t max_buffer_size = strlen(name) + strlen(TOSTRING(DYNLIB_MAX_MAJOR_VERSION)) +
+    size_t max_buffer_size = strlen(name) + strlen(TOSTRING(DYNLIB_MAX_VERSION)) +
                              strlen(TOSTRING(DYNLIB_MAX_MINOR_VERSION)) + strlen(".") + strlen(".") +
```
MsDerekMa: Should this be the string length of "0"?  Or of "DYNLIB_MAX_VERSION"?  I thought DYNLIB_MAX_MINOR_VERSION went away.

include/k4ainternal/k4aplugin.h
```diff
@@ -23,13 +23,16 @@ extern "C" {
 #endif
 
 /**
- * Current Version of the k4A Depth Engine Plugin API
+ * Current Version of the Azure Kinect Depth Engine Plugin
```
MsDerekMa: Should we call this the Azure Kinect Depth Engine Plugin Interface?

include/k4ainternal/k4aplugin.h
```diff
@@ -23,13 +23,16 @@ extern "C" {
...
  *
- * \remarks The MAJOR version must be updated to denote any breaking change
- * that would cause an older SDK to not be able to use this plugin.
+ * \remarks
+ * When the depth engine plugin is updated this version should be increased.
```
MsDerekMa: Sure this be more specific that when the interface it updated that this should be changed.

src/deloader/deloader.cpp
```diff
@@ -45,20 +45,11 @@ static bool verify_plugin(const k4a_plugin_t *plugin)
...
 
-    // Major versions must match
-    if (plugin->version.major != K4A_PLUGIN_MAJOR_VERSION)
-    {
-        LOG_ERROR("Plugin Major Version Mismatch. Expected %u. Found %u",
-                  K4A_PLUGIN_MAJOR_VERSION,
-                  plugin->version.major);
-        return false;
-    }
-
```
MsDerekMa: Our goal is that the number in the file name is our "guarantee" that the file is compatible?

src/dynlib/dynlib_linux.c
wes-b: I made a mistake elsewhere and though Linux was not using minor version in the file name, I will fix this too

src/deloader/deloader.cpp
wes-b: yes

include/k4ainternal/dynlib.h
```diff
@@ -20,39 +20,30 @@ extern "C" {
...
  *
  * \remarks
  * The version information should be encoded in the filename of the dynamic
- * library. This encoding will be different for Windows versus Linux. For
- * Windows the dynamic library name is "<name>_<major_ver>_<minor_ver>.dll".
+ * library being loaded. This encoding will be different for Windows versus Linux. For
+ * Windows the dynamic library name is "<name>_<version>_0.dll".
  * For Linux, the dynamic library name is
- * "lib<name>.so.<major_ver>.<minor_ver>".
+ * "lib<name>.so.<version>".
```
schultetwin1: Just a suggestion from what we talked about earlier

```suggestion
 * "lib<name>.so.<version>.0".
```

src/dynlib/dynlib_linux.c
```diff
@@ -11,6 +11,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <ctype.h>
+#define __USE_GNU
```
schultetwin1: I think we want to use `_GNU_SOURCE`

https://stackoverflow.com/questions/7296963/gnu-source-and-use-gnu

```suggestion
#define _GNU_SOURCE
```

src/dynlib/dynlib_linux.c
```diff
@@ -21,13 +22,13 @@ typedef struct _dynlib_context_t
...
 K4A_DECLARE_CONTEXT(dynlib_t, dynlib_context_t);
-static char *generate_file_name(const char *name, uint32_t major_ver, uint32_t minor_ver)
+static char *generate_file_name(const char *name, uint32_t version)
 {
     const char *lib_prefix = "lib";
     const char *lib_suffix = "so";
-    size_t max_buffer_size = strlen(name) + strlen(TOSTRING(DYNLIB_MAX_MAJOR_VERSION)) +
-                             strlen(TOSTRING(DYNLIB_MAX_MINOR_VERSION)) + strlen(".") + strlen(".") +
-                             strlen(lib_suffix) + strlen(lib_prefix) + 1;
+    // Format of the depth engine name is: libdepthengine.so.2.0
```
schultetwin1: What about making the format less dependent on the current version?

```suggestion
    // Format of the depth engine name is: libdepthengine.so.<K4A_PLUGIN_VERSION>.0
    //                                 ex: libdepthengine.so.2.0
```

src/dynlib/dynlib_windows.c
```diff
@@ -21,10 +21,10 @@ typedef struct _dynlib_context_t
 
 K4A_DECLARE_CONTEXT(dynlib_t, dynlib_context_t);
 
-static char *generate_file_name(const char *name, uint32_t major_ver, uint32_t minor_ver)
+static char *generate_file_name(const char *name, uint32_t version)
 {
-    size_t max_buffer_size = strlen(name) + strlen(TOSTRING(DYNLIB_MAX_MAJOR_VERSION)) +
-                             strlen(TOSTRING(DYNLIB_MAX_MINOR_VERSION)) + strlen("_") + strlen("_") + 1;
+    // Format of the depth engine name is: depthengine_2_0
```
schultetwin1: Same comment here about the using a specific version

PR #905, opened 2019-11-13
Allows recorder to set exposure based on desired exposure setting.

### LINKED ISSUE #904: K4aRecorder should all exposure to be set in microseconds
**Describe the bug**
The Tool should allow exposure to be set in micro seconds.
The tool should also allow gain to be adjusted.

wes-b:
- Allows user to enter 2-200,000us for exposure, this is in addition to continuing to allow -11 to 1 for the existing MFC implementation.
- Allows user to specify gain


wes-b:
@xthexder thanks for the quick review

PR #912, opened 2019-11-15
C++ for recorder

### LINKED ISSUE #911: CPP headers need test coverage

wes-b:
- Minor interface change for recorder.hpp (Potential Breaking Change for recorder.hpp users)
- Minor interface change to all C++ bool operators. == and != no longer allowed on device and playback classes
- Adds tests for k4a.hpp, playback.hpp, recorder.hpp,


include/k4arecord/playback.hpp
```diff
@@ -312,6 +547,77 @@ class playback
...
+        }
+
+        throw error("Failed to get previous data block!");
+    }
+
+    /** Get the attachment block from the recording.
+     * Returns true if the attachment was available, false if it was not found.
+     * Throws error on failure.
+     *
+     * \sa k4a_playback_get_previous_data_block
```
xthexder: k4a_playback_get_attachment

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,453 @@
...
+    kinect = kinect.open(0);
+
+    {
+        ASSERT_LE((uint32_t)1, device::get_installed_count());
+    }
+
+    {
+        k4a_hardware_version_t version = kinect.get_version();
+        ASSERT_GE(version.rgb.major, (uint32_t)1);
+        ASSERT_GE(version.rgb.minor, (uint32_t)6);
```
xthexder: This will break if we ever upgrade to a 2.0 firmware.
Maybe that's not important and we can just update the test then.

include/k4arecord/playback.hpp
```diff
@@ -14,11 +14,218 @@
...
+        if (m_handle == nullptr)
+        {
+            throw error("Failed std::make_shared for data_block");
+        }
+    }
+
+    /** Creates a shallow copy of another data_block
+     */
+    data_block(const data_block &other) noexcept : m_handle(other.m_handle)
+    {
```
xthexder: I think this will always result in an invalid handle.
It would be good to add some tests for these constructors as well.

include/k4a/k4a.hpp
```diff
@@ -109,7 +109,7 @@ class image
         reset();
     }
 
-    /** Sets image to a shallow copy of other
+    /** Sets image to a shallow copy of the other
```
williampMSFT: >the other [](start = 40, length = 9)

I think either of these is correct:
"shallow copy of other" (treating "other" as a reference to the actual name of the parameter), or
"shallow copy of the other image" (treating "other" as an adjective describing the image)

No strong feeling on which one to go with, but this way reads like it's missing a word; ditto for line 438

include/k4arecord/playback.hpp
```diff
@@ -14,11 +14,218 @@
 #include <chrono>
 #include <string>
 #include <vector>
+#include <memory>
```
williampMSFT: nit: sort includes alphabetically for easier merge (might not matter on this product but is probably good practice)

include/k4arecord/record.hpp
```diff
@@ -102,11 +130,11 @@ class record
      *
      * \sa k4a_record_add_tag
      */
-    void add_tag(const char *name, const char *value)
+    void add_tag(const char *name, const std::string &value)
```
williampMSFT: I'm not sure I understand the rationale behind this change.  I think this just makes it so you have to wrap a C string in a STL string, but we're not actually taking ownership of that string, are we?  I don't object to adding an overload to take a std::string, but I think we probably want to keep the C-string overload, no?

Also it seems a little weird that we're mixing and matching C and C++ strings here

include/k4arecord/playback.hpp
williampMSFT: Yeah, I think you need to remove the m_handle.reset()

---
In reply to: [347034431](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/912#discussion_r347034431) [](ancestors = 347034431)

include/k4arecord/playback.hpp
```diff
@@ -14,11 +14,218 @@
...
+    /** Moves another data_block into a new data_block
+     */
+    data_block(data_block &&other) noexcept : m_handle(other.m_handle)
+    {
+        other.m_handle = nullptr;
+    }
+
+    ~data_block()
+    {
+        reset();
```
williampMSFT: >reset(); [](start = 8, length = 8)

This is not strictly necessary; std::shared_ptr/std::unique_ptr do this automatically in their destructors

include/k4arecord/playback.hpp
```diff
@@ -18,7 +18,213 @@
...
+            return m_handle != nullptr;
+        }
+
+    private:
+        k4a_playback_data_block_t m_handle;
+    };
+
+public:
+    /** Creates a data_block from a k4a_playback_data_block_t
+     * Takes ownership of the handle, you should not call k4a_playback_data_block_release? on the handle after
```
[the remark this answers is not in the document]
wes-b: Thanks, good catch

include/k4arecord/playback.hpp
```diff
@@ -14,11 +14,218 @@
...
+     * giving it to the data_block; the data_block will take care of that.
+     */
+    data_block(k4a_playback_data_block_t handle = nullptr)
+    {
+        m_handle = std::make_shared<data_block_ref>(handle);
+        if (m_handle == nullptr)
```
williampMSFT: >if (m_handle == nullptr) [](start = 8, length = 24)

Pretty sure this can't ever happen - the only way for this to happen is if data_block_ref's constructor throws (in which case we won't get this far because a k4a::error was thrown) or if operator new throws (in which case we won't get this far because a std::bad_alloc has been thrown)

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
wes-b: Yea, we will worry about this one when there is a 2.0 or greater.

include/k4arecord/playback.hpp
```diff
@@ -14,11 +14,218 @@
...
+/** \class data_block playback.hpp <k4arecord/playback.hpp>
+ * Wrapper for \ref k4a_playback_data_block_t
+ *
+ * \sa k4a_playback_data_block_t
+ */
+class data_block
+{
+private:
+    class data_block_ref
+    {
```
williampMSFT: I might be missing something, but it looks like we have two classes wrapping this to add refcounting semantics to a type that doesn't have them natively - why are we doing that?  We didn't for k4a_device_t; there, we just removed the copy constructors and only left a move constructor.

include/k4arecord/playback.hpp
```diff
@@ -72,6 +279,34 @@ class playback
...
+        return m_handle != other.m_handle;
+    }
+
+    /** Returns true if the playback object is valid, false otherwise
+     */
+    bool operator!=(std::nullptr_t) const noexcept
+    {
+        return m_handle != nullptr;
+    }
```
williampMSFT: I'm not sure these work with the semantics of k4a::playback - it's not a pseudo-pointer type like k4a::image, we delete the copy constructor so two playbacks should never be equal and we don't allow assigning nullptr to the playback, so allowing comparison to nullptr seems really weird

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,453 @@
...
+        cap1.set_temperature_c(0.0f);
+        cap2.set_temperature_c(0.0f);
+
+        {
+            capture shallow_copy;
+            capture deep_copy;
+            shallow_copy = cap1;               // test = operator
+            ASSERT_TRUE(shallow_copy == cap1); // test == operator
+
+            deep_copy = std::move(cap1);       // test = (&&) operator
```
williampMSFT: >deep_copy = std::move(cap1) [](start = 12, length = 27)

This isn't a deep copy, it's a move - the object is never copied, it's just that the ownership of the handle changes

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,453 @@
...
+
+    for (int x = 0; x < 100; x++)
+    {
+        capture capture;
+        k4a_imu_sample_t imu;
+        if (kinect.get_capture(&capture, std::chrono::milliseconds(1000)))
+        {
+            recorder.write_capture(capture);
+        }
+        while (kinect.get_imu_sample(&imu, std::chrono::milliseconds(0)))
```
williampMSFT: >while [](start = 8, length = 5)

I think this can hang forever if the system somehow is running slow enough that it can't keep up with incoming IMU samples - we might want a max here so the test agent doesn't hang indefinitely if some other process on the test agent is hammering the processor

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,453 @@
...
+    ASSERT_FALSE(pb == pb_empty);
+
+    // Deep copy
+    playback pback = std::move(pb);
+    ASSERT_TRUE(pback != nullptr);
+    ASSERT_TRUE(pb == nullptr);
+
+    std::vector<uint8_t> raw_cal = pback.get_raw_calibration();
+    std::cout << "calibration is : ";
+    for (std::vector<uint8_t>::const_iterator i = raw_cal.begin(); i != raw_cal.end(); ++i)
```
williampMSFT: >std::vector<uint8_t>::const_iterator i = raw_cal.begin(); i != raw_cal.end(); ++i [](start = 9, length = 81)

This is fine, but FYI C++ has foreach loops as of C++11 - you can write this more concisely as:
for (const uint8_t& char : raw_cal)
{
    std::cout << char;
}

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,453 @@
...
+        capture.reset();
+    }
+    recorder.flush();
+    kinect.stop_cameras();
+    kinect.stop_imu();
+}
+
+TEST_F(k4a_cpp_ft, playback)
+{
+    playback pb = playback::open("./k4a_cpp_ft.mkv");
```
williampMSFT: >k4a_cpp_ft.mkv [](start = 36, length = 14)

I don't see this file in this PR - did it get added?

include/k4arecord/record.hpp
```diff
@@ -67,13 +67,41 @@ class record
...
+    {
+        return m_handle != other.m_handle;
+    }
+
+    /** Returns true if the record object is valid, false otherwise
+     */
+    bool operator!=(std::nullptr_t) const noexcept
+    {
+        return m_handle != nullptr;
+    }
```
williampMSFT: I'm not sure these work with the semantics of k4a::record - it's not a pseudo-pointer type like k4a::image, we delete the copy constructor so two records should never be equal and we don't allow assigning nullptr to the playback, so allowing comparison to nullptr seems really weird

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
xthexder: I think it's supposed to be created by the record test.

On that note, it should be deleted in test suite teardown, similar to how the c tests operate.

include/k4arecord/playback.hpp
wes-b: Good question, I thought if I deleted the assignment operator I would loose the ability to pass the value to function. I like this option much better. Thank you!

include/k4arecord/playback.hpp
wes-b: It is the same, just not ref counted, exactly the same as 'device'. Seems like if we had the bool operator then we should allow for == != nullptr? no?

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
wes-b: you are right, I updated all the references to deep copy

include/k4arecord/playback.hpp
williampMSFT: The device class also doesn't have comparison operators for the same reason :)  The problem is you actually can't have 2 playbacks/records/devices wrapping the same underling C playback/record/device handles, so comparing to anything doesn't make sense.  We don't provide a copy constructor or copy operator= for those types because if we did, the second wrapper would segfault in its destructor because it's trying to free memory that's already been freed by the first one's destructor.  I suppose you could define just operator==(nullptr_t), but that's also super weird because it's not really behaving like a refcounted pointer in the same way a k4a::image is, we're just randomly using ==(nullptr_t) as an is_valid() function.  operator bool is idiomatic for "is this thing valid" even on non-pointerish types (e.g. std::ostream/std::istream do this), but they don't allow pointer type comparisons either

---
In reply to: [347614993](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/pull/912#discussion_r347614993) [](ancestors = 347614993)

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
wes-b: Thanks!

include/k4arecord/record.hpp
wes-b: It is also common in C to check handles against NULL, so this seems fine to me

include/k4arecord/playback.hpp
```diff
@@ -14,11 +14,218 @@
...
+        void reset()
+        {
+            if (m_handle != nullptr)
+            {
+                k4a_playback_data_block_release(m_handle);
+                m_handle = nullptr;
+            }
+        }
+
+        k4a_playback_data_block_t get_handle()
```
wes-b: remove this function

include/k4ainternal/common.h
```diff
@@ -140,6 +140,52 @@ inline static bool k4a_convert_depth_mode_to_width_height(k4a_depth_mode_t mode,
     return true;
 }
 
+inline static bool k4a_is_version_greater_or_equal(k4a_version_t *fw_version_l, k4a_version_t *fw_version_r)
```
wes-b: Wrong change

include/k4arecord/playback.hpp
```diff
@@ -12,13 +12,115 @@
...
+    {
+        reset();
+    }
+
+    /** Moves another data_block into this data_block; other is set to invalid
+     */
+    data_block &operator=(data_block &&other) noexcept
+    {
+        reset();
+        if (this != &other)
```
williampMSFT: I think you want the reset inside the if block

include/k4arecord/playback.hpp
```diff
@@ -12,13 +12,115 @@
...
+        }
+        return *this;
+    }
+
+    /** Invalidates this data_block
+     */
+    data_block &operator=(std::nullptr_t) noexcept
+    {
+        return *this;
+    }
```
williampMSFT: This doesn't seem to do anything - I don't think we want this to have pointer semantics at all, so I'd just delete this function entirely

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,511 @@
...
+    playback pback;
+    ASSERT_TRUE(pb);     // bool operation
+    ASSERT_FALSE(pback); // bool operation
+    pback = std::move(pb);
+    ASSERT_FALSE(pb);   // bool operation
+    ASSERT_TRUE(pback); // bool operation
+
+    std::vector<uint8_t> raw_cal = pback.get_raw_calibration();
+    std::cout << "calibration is : ";
+    for (const uint8_t data_char : raw_cal)
```
williampMSFT: >d [](start = 23, length = 1)

nit: you're looping over these by value, not by reference, so it's actually copying each char into a temporary here.  This is test code so it probably doesn't really matter, but you probably want an & here

tests/RecordTests/FunctionalTest/k4a_cpp_ft.cpp
```diff
@@ -0,0 +1,511 @@
...
+{
+    return k4a_test_common_main(argc, argv);
+}
+
+class k4a_cpp_ft : public ::testing::Test
+{
+public:
+    virtual void SetUp()
+    {
+        std::remove("*.mkv"); // remove old test files in the event of a crash
```
williampMSFT: >std::remove("*.mkv") [](start = 8, length = 20)

I'm pretty sure this doesn't actually work - I don't think wildcards are supported using std::remove.  I think you need to call std::remove on every file you want to delete.  Unfortunately, std::filesystem is a C++17 feature, and that's the only built-in portable way to enumerate files I'm aware of, so I don't think we have access to the stuff to enumerate files in the K4A project (it's on C++11, I think?), so you might just have to hardcode them here :(

include/k4a/k4a.hpp
```diff
@@ -176,7 +176,14 @@ class image
...
      */
-    operator bool() const noexcept
+    explicit operator bool() const noexcept
+    {
+        return m_handle != nullptr;
+    }
+
+    /** Returns true if the image is valid, false otherwise
+     */
+    bool good() const noexcept
```
xthexder: I think calling this `valid()` might be more self-explanatory.

include/k4arecord/playback.hpp
```diff
@@ -18,7 +18,108 @@
...
+    // No Copies allowed
+    data_block(const data_block &) = delete;
+    data_block &operator=(const data_block &) = delete;
+
+    /** Moves another data_block into a new data_block
+     */
+    data_block(data_block &&other) noexcept : m_handle(other.m_handle)
+    {
+        reset();
+        m_handle = other.m_handle;
```
xthexder: You're doing this twice now, and I think potentially calling `k4a_playback_data_block_release` on the handle.

I think just removing ` : m_handle(other.m_handle)` would be fine.

xthexder: Calling `reset()` in the constructor shouldn't be required either, since it should always be a new object.

include/k4arecord/playback.hpp
```diff
@@ -312,6 +420,77 @@ class playback
...
+     * Returns true if the attachment was available, false if it was not found.
+     * Throws error on failure.
+     *
+     * \sa k4a_playback_get_attachment
+     */
+    bool get_attachment(const char *attachment, std::vector<uint8_t> *data)
+    {
+        size_t data_size = 0;
+        k4a_buffer_result_t result = k4a_playback_get_attachment(m_handle, attachment, nullptr, &data_size);
+        if (result == K4A_BUFFER_RESULT_TOO_SMALL)
```
xthexder: If the provided buffer is large enough, this will always return false.

xthexder: Nevermind, I read this wrong.

PR #929, opened 2019-11-20
Updates for FW 1.6.107078014

wes-b:
- Modifies Color Gain test as default changes
- Makes FW version checking common code common.h
  - depth.c updated to use common code
- Making expected capture timestamp synchronization tolerance to be based on FPS.
  - Updated throughput test and multidevice tests
- Updated multidevice tests to run in looping mode better and tolerant to 1 off errors.
- Fix a bug in capturesync.c where dropped frame count was not zeroed out.


tests/UnitTests/utcommon/inc/utcommon.h
```diff
@@ -29,6 +29,8 @@ void k4a_unittest_init_logging_with_processid();
 #endif
 
 int k4a_test_common_main(int argc, char **argv);
+
+int64_t k4a_unittest_get_max_sync_delay(k4a_fps_t fps);
```
schultetwin1: nit: Can you add either a comment of change this function name to include units of `max_delay`?

```suggestion
int64_t k4a_unittest_get_max_sync_delay_ms(k4a_fps_t fps);
```

tests/UnitTests/utcommon/utcommon.cpp
```diff
@@ -107,4 +107,25 @@ int k4a_test_common_main(int argc, char **argv)
...
+        break;
+    case K4A_FRAMES_PER_SECOND_15:
+        max_delay = 220;
+        break;
+    case K4A_FRAMES_PER_SECOND_30:
+        max_delay = 110;
+        break;
+    default:
+        max_delay = 0;
+        break;
```
schultetwin1: I believe we might want to remove `default` here. This `switch` statement has all possible enumerations of `k4a_fps_t`. Therefore it should compile without `default`. If someone were to add a new enumeration to `k4a_fps_t` then they would get a compiler break for this `switch` statement (which we probably want) and they would have to add the max delay here.

tests/multidevice/multidevice.cpp
```diff
@@ -63,9 +60,25 @@ class multidevice_sync_ft : public ::testing::Test
...
+            m_master = nullptr;
+        }
+        if (m_subordinate != nullptr)
+        {
+            k4a_device_close(m_subordinate);
+            m_subordinate = nullptr;
+        }
+    }
+    k4a_device_t m_master = nullptr;
+    k4a_device_t m_subordinate = nullptr;
```
schultetwin1: I don't think we want to add the types here to `m_master` and `m_subordinate`. This is just declaring unused local variables right?

tests/UnitTests/utcommon/utcommon.cpp
wes-b: Sure

tests/multidevice/multidevice.cpp
wes-b: No, this is the class that gets constructed first, and setup and teardown are calls that run at the beginning and end of the program. Tear down runs in the event of a crash so it can help ensure cleanup.

schultetwin1: Let me ask a different way, do we need to have the type of `m_master` and `m_subordinate` or could we just re-write these two lines as (notice the removal of `k4a_device_t` from each line):

```c++
m_master = nullptr;
m_subordinate = nullptr;
```

I _think_ (but am not sure) that adding the `k4a_device_t` in front of `m_master` and `m_subordinate` has made it so that we are not referring to local variables instead of member variables.

schultetwin1: Wes talked  with me about this, I thought this was still in a function defintion, but it's obviously in a class definition which makes more sense.

PR #956, opened 2019-12-05
Reducing Rational 6KT error message rate

### LINKED ISSUE #950: The 6KT deprecation info is overly verbose
**Describe the bug**
When use an EV Azure Kinect sensor, the calibration format was 6KT, which is supported by the SDK. However, after sensor SDK 1.3, it will overly verbose printing a critical error: "Rational 6KT is deprecated (only supported early internal devices). Please replace your Azure Kinect with a retail device." It makes the old device / recordings almost unusable without disabling any error loggings. 

The logging can be found here: https://github.com/microsoft/Azure-Kinect-Sensor-SDK/blob/6762f44a13061fce5c239fecdaeb3726a53774ce/src/transformation/intrinsic_transformation.c#L37 

**To Reproduce**
Run sample code with an EV device. 

**Expected behavior**
Make it as a warning instead of critical error. And only print it out once. 

 - SDK Version:  1.3

wes-b:
- Updates warning to only fire 1 time per session instead of many

### I tested changes on: 
- [x] Windows
- [ ] Linux

Manually tested to confirm change.

### DIFF
--- a/src/transformation/intrinsic_transformation.c
+++ b/src/transformation/intrinsic_transformation.c
@@ -6,6 +6,11 @@
 
 #include <float.h>
 
+// We don't like globals if we can help it. This one is for reducing critical logging noise when recorded files are used
+// with Rational 6KT calibration. Production devices never had this calibration but recordings were made with this
+// calibration. So we fire the warning 1 time instead of every time a transformation call is made
+static int g_deprecated_6kt_message_fired = false;
+
 static k4a_result_t transformation_project_internal(const k4a_calibration_camera_t *camera_calibration,
                                                     const float xy[2],
                                                     float uv[2],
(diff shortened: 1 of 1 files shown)


src/transformation/intrinsic_transformation.c
```diff
@@ -248,8 +255,10 @@ static k4a_result_t transformation_unproject_internal(const k4a_calibration_came
         return K4A_RESULT_FAILED;
     }
 
-    if (camera_calibration->intrinsics.type == K4A_CALIBRATION_LENS_DISTORTION_MODEL_RATIONAL_6KT)
+    if (camera_calibration->intrinsics.type == K4A_CALIBRATION_LENS_DISTORTION_MODEL_RATIONAL_6KT &&
+        g_deprecated_6kt_message_fired == false)
     {
+        g_deprecated_6kt_message_fired = true;
         LOG_CRITICAL("Rational 6KT is deprecated (only supported early internal devices). Please replace your Azure "
```
yijiew: Do we want to log the information as Critical level? Maybe warning will be enough?

wes-b: The issue is still critical for customers, so it should stay critical. If the calibration turns up then a production device is needed.

wes-b:
Thanks for the quick review

src/transformation/intrinsic_transformation.c
yijiew: I see. It makes sense :)

PR #962, opened 2019-12-09
K4aRecorder support recording BGRA32 stream as MJPEG

### LINKED ISSUE #404: Support recording K4A_IMAGE_FORMAT_COLOR_BGRA32 image type
BGRA color image type is convenient for computer vision algorithms. It is very likely for a customer to config the device/playback to provide the K4A_IMAGE_FORMAT_COLOR_BGRA32 type. However, the record API doesn't support recording this type. 

Record API supports recording K4A_IMAGE_FORMAT_COLOR_BGRA32 image type. 

**Describe alternatives you've considered**
Or provide a convenient API for the user to convert the K4A_IMAGE_FORMAT_COLOR_BGRA32 to other recording-supported types like: K4A_IMAGE_FORMAT_COLOR_NV12, K4A_IMAGE_FORMAT_COLOR_YUY2...

**Additional context**
None

wes-b:
- Recorder will convert BGRA32 frames to MJPEG and create a recording with an MJPEG stream
- Recorder already supports reading MJPEG streams as BGRA32

I also played the resulting files in K4aViewer to confirm they were compress as expected.

### DIFF
--- a/include/k4ainternal/matroska_write.h
+++ b/include/k4ainternal/matroska_write.h
@@ -99,6 +99,8 @@ typedef struct _k4a_record_context_t
     std::mutex writer_lock;
 
     bool header_written, first_cluster_written;
+
+    void * /*tjhandle*/ compress_handle;
 } k4a_record_context_t;
 
 K4A_DECLARE_CONTEXT(k4a_record_t, k4a_record_context_t);
--- a/src/record/sdk/record.cpp
+++ b/src/record/sdk/record.cpp
@@ -11,18 +11,24 @@
 #include <k4ainternal/logging.h>
 #include <k4ainternal/common.h>
 
+#include <turbojpeg.h>
+
+// With this quality the generated MKV file from a BGRA stream is approximately the same as the MJPEG stream
+const int JPEG_QUALITY = 94;
+
 using namespace k4arecord;
 using namespace LIBMATROSKA_NAMESPACE;
 
 k4a_result_t k4a_record_create(const char *path,
                                k4a_device_t device,
-                               const k4a_device_configuration_t device_config,
+                               const k4a_device_configuration_t user_device_config,
                                k4a_record_t *recording_handle)
 {
     RETURN_VALUE_IF_ARG(K4A_RESULT_FAILED, path == NULL);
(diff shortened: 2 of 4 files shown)


xthexder:
I'm still not sure how I feel about this solution, since we'll be blocking the `write_capture` function while compressing the image, and while probably minimal, there will be some reduction in image quality due to decompressing and recompressing the MJPEG images.

Unfortunately I'm not sure what a better solution would be, since it would require some fairly major changes to get at the raw MJPEG images off the camera when in BGRA mode.
There may be some way to move this compression to the writing thread so the caller isn't blocked, but that's not a very simple change either.

yijiew:
@xthexder : yours is a very good concern. Depdending on the MJPEG algorithm we use, compressing to MJPEG twice might cause too much data loss.  Actually, instead of compressing to MJPEG, we could consider using NV12? Because it is a close form conversion. I know for sure NV12->RGBA->NV12 should be identical.

wes-b:
We discussed this offline. Instead of decompressing from native MJPEG to BGRA32 in the SDK and then recompressing to MJPEGin the recorder, we will instead refer users to stream in MJPEG, record in MJPEG, and provide sample code to convert MJPEG to BGRA32. We will also enable BGRA32 recording for users with PC's and file systems that can handle that demand.

UnaNancyOwen:
@wes-b
I think it is good this option should be provided to users. (It would be even better if other formats could be supported too.)
I understand that there is degradation in image quality by decoded image is re-compressed. It should be documented, or warning.
If developers want to ensure data identity, MotionJPEG (or lossless format such as NV12) should be used. But, developers may not want data identity. That will depend on the use case.
BGRA32 format support is useful when recording with displaying. It can reduce self-implementation of decoding process of MotionJPEG from user code.

Please re-consider to merge this feature. Thanks,

wes-b:
@UnaNancyOwen I am curious about what you like about this PR that you could not do with the new plan. With this PR you can not record BGRA32. Instead we compress it and store as MJPEG. With sample code to convert from MJPEG to BGRA32 it seems like it provides a solution for all BGRA32/MJPEG users.

PR #1059, opened 2020-02-04
Arm64 work

schultetwin1:
- Adds support for CPU detection by CMake
- Removes SSE instructions from ARM builds
- Adds CMake ARM toolchains files
- Updates the docker image to support cross compiling

### I tested changes on:
- [ ] Windows
- [x] Linux

This is currently a draft to get our CI system up and running.


src/transformation/rgbz.c
```diff
@@ -8,9 +8,12 @@
 #include <limits.h>
 #include <math.h>
 
+#if defined(__amd64__) || defined(_M_AMD64) || defined(__i386__) || defined(_M_X86)
```
wes-b:
```suggestion
#ifdef __SSE4_1__
```

src/transformation/rgbz.c
```diff
@@ -1055,42 +1058,45 @@ k4a_buffer_result_t transformation_color_image_to_depth_camera_internal(
     return K4A_BUFFER_RESULT_SUCCEEDED;
 }
 
+#if !defined(K4A_USING_SSE)
```
wes-b:
```suggestion
#ifndef __SSE4_1__
```

src/transformation/rgbz.c
schultetwin1: Does that work for MSVC as well?

From here it looks like it does not:
https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

wes-b: no

wes-b:
Work has been completed on a feature branch instead of this branch

PR #1175, opened 2020-04-14
Night builds managed by Azure-Pipelines.yml

wes-b:
Switching away from the ADO UI

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@ name: K4A-SDK-$(SourceBranchName)-$(Date:yyyyMMdd)-$(Rev:rrr)
 
 schedules:
 - cron: "*/30 08-13 * * *"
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -3,6 +3,68 @@
 
 name: K4A-SDK-$(SourceBranchName)-$(Date:yyyyMMdd)-$(Rev:rrr)
 
+schedules:
+- cron: "30 8 * * *"
```
schultetwin1, line 7: I think we can simplify all this into one cron schedule

```suggestion
- cron: "*/30 08-13 * * *"
```

Check out this website for more info on cron expressions

https://crontab.guru/#*/30_08-13_*_*_*

wes-b: BTW, thanks for the  tip, that cleaned things up a lot! I forgot to mention that before I submitted last night

schultetwin1: Just trying to turn your into a Linux SysAdmin 😄

PR #1177, opened 2020-04-14
SIMD: treat remainder correctly

tomoaki0705:
## Fixes #

If the input image size element is not multiple of 8, small portion of data get unprocessed.
I'm not sure if there is a guarantee that the input size become multiple of 8.

- apply naive process if there is a remainder

### I tested changes on: 
- [ ] Windows
- [x] Linux

### DIFF
--- a/src/transformation/rgbz.c
+++ b/src/transformation/rgbz.c
@@ -1083,12 +1083,12 @@ k4a_buffer_result_t transformation_color_image_to_depth_camera_internal(
     return K4A_BUFFER_RESULT_SUCCEEDED;
 }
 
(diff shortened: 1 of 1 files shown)


wes-b:
Are you using an image size that isn't a multiple of 8? If so you mind sharing?

tomoaki0705:
No, I'm not.
The motivation of this PR is my general approach when using SIMD.
If the image size is guaranteed multiple of 8, this PR is not needed.

wes-b:
This raises a good point. At a minimum we need to add a check

PR #1203, opened 2020-05-04
Update azure-pipelines.yml

wes-b:
Triggering Automation builds


wes-b:
The UI was controlling schedule when this was last evaluated, so it didn't take. Now UI scheduling is removed we need a commit to re-eval YAML based schedule.

schultetwin1:
> The UI was controlling schedule when this was last evaluated, so it didn't take. Now UI scheduling is removed we need a commit to re-eval YAML based schedule.

Huh, that's interesting behavior. I would've expected the YAML to override the UI.

PR #1222, opened 2020-05-21
Reduce usage of build pools

schultetwin1:
We are currently using over 30 machines in parallel. This will reduce
our usage down to 15 machines by staggering the master and develop
builds as well as removing some build flavors we don't really care
about.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,11 +4,18 @@
 name: K4A-SDK-$(SourceBranchName)-$(Date:yyyyMMdd)-$(Rev:rrr)
 
 schedules:
-- cron: "*/30 08-13 * * *"
-  displayName: 'Daily Builds starting at 8:00AM UTC (12:00AM PST) every 30 min, last build 13:30AM (5:30AM PST)'
+# Every 30 minutes from 08:00 UTC (00:00 PST/01:00 PDT) to 10:00 UTC (02:00 PST/03:00 PDT)
+- cron: "*/30 08-10 * * *"
+  displayName: 'Nightly Builds - Develop'
   branches:
     include:
-    - master
+    - develop
+  always: true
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -4,11 +4,18 @@
...
-  displayName: 'Daily Builds starting at 8:00AM UTC (12:00AM PST) every 30 min, last build 13:30AM (5:30AM PST)'
+# Every 30 minutes from 08:00 UTC (00:00 PST/01:00 PDT) to 10:00 UTC (02:00 PST/03:00 PDT)
+- cron: "*/30 08-10 * * *"
+  displayName: 'Nightly Builds - Develop'
+  branches:
+    include:
+    - develop
+  always: true
+# Every 30 minutes from 11:00 UTC (03:00 PST/04:00 PDT) to 13:00 UTC (05:00 PST/06:00 PDT)
+- cron: "*/30 10-13 * * *"
```
wes-b, line 15:
```suggestion
- cron: "*/30 11-13 * * *"
```

azure-pipelines.yml
```diff
@@ -4,11 +4,18 @@
...
+  displayName: 'Nightly Builds - Master'
   branches:
     include:
-    - master
     - develop
```
wes-b, line 19:
```suggestion
    - master
```

azure-pipelines.yml
```diff
@@ -89,22 +96,6 @@ jobs:
         MSBuildConfiguration: 'Release'
         vmImage: 'vs2017-win2016'
         UsesOpenCV: 'FALSE'
-      VS2017_x86-msvc_debug_msbuild:
```
wes-b, line 92: We should validate these changes with the packaging build too, We are getting ready to hand this over it would be nice to know this won't be a problem for them in several month when a new release gets planned

schultetwin1: All of these builds that we no longer run are not run on the packaging build either. The packaging build just runs:

32-bit and 64-bit builds using VS2017 and a Ninja generator

PR #1608, opened 2021-05-25
fix search for python 3

### LINKED ISSUE #1583: `scripts/verify-windows.ps1` fails to find installed python3
` scripts/verify-windows.ps1` does not find python3 that has been installed by Windows itself. The script reports the error

```
C:\njs\Azure-Kinect-Sensor-SDK\scripts\verify-windows.ps1 : Unable to find python 3. Please install python3 and add it
to your path
At line:1 char:1
+ .\verify-windows.ps1
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,verify-windows.ps1
```

### Setup

* Microsoft Windows [Version 10.0.19042.928]
* Python3 installed direct from the Windows App Store. Windows key, store, search for python3, install.
* The remaining components described at https://github.com/microsoft/Azure-Kinect-Sensor-SDK/blob/develop/docs/building.md
* git clone this repo to a directory on your hard drive

### Repro

1. Run `scripts/verify-windows.ps1`

### Result

```
C:\njs\Azure-Kinect-Sensor-SDK\scripts\verify-windows.ps1 : Unable to find python 3. Please install python3 and add it
to your path
At line:1 char:1
+ .\verify-windows.ps1
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,verify-windows.ps1
```

### Expected

No error and the successful output "Machine setup. Please use Visual Studio Developer Command prompt to build."

### Notes

If I run the command of line 86 of the verify script `Get-Command -All -Name "python"` I get the result

```
PS C:\njs\Azure-Kinect-Sensor-SDK\scripts> Get-Command -All -Name "python"

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     python.exe                                         0.0.0.0    C:\Users\dale\AppData\Local\Microsoft\WindowsApps\python.exe
```

and if at that same ps prompt I type that long path, or python, or python3...then all three methods run python 3...

```
Python 3.8.9 (tags/v3.8.9:a743f81, Apr  2 2021, 11:10:41) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

This suggests that the verify script's method of version checking is not sufficient to detect Windows Store installed python. Naturally, the Windows Store is....desired to also be supported. This is likely related to the Windows Store and App Execution Aliases.

It is not possible to add the Windows Store python manually to the PATH. That location is typically cryptic and not user maintainable. For example, on my computer it is `C:\Users\dale\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0`

diablodale:
- fixes microsoft/Azure-Kinect-Sensor-SDK#1583
- change is to a build verification script, not to SDK code/api itself

- method-wide check for python3 on path and available to execute

- [n/a] I ran the [performance tests](https://github.com/Microsoft/Azure-Kinect-Sensor-SDK/blob/develop/docs/testing.md) with a device

### I tested changes on:
- [X] Windows
- [ ] Linux

via projects automated test scripts. And used in my my project.

### DIFF
--- a/scripts/verify-windows.ps1
+++ b/scripts/verify-windows.ps1
@@ -82,7 +82,7 @@ if (-not (Get-Command -Name "tar")) {
 }
 
 # Check for python 3
-if (-not (Get-Command -All -Name "python" | Where-Object { $_.Version -ge [Version]"3.0" }))
+if (-not (Get-Command -All -Name "python3"))
 {
     Write-Error "Unable to find python 3. Please install python3 and add it to your path"
 }


jaygullapalli:
This check was created to support traditional python installer from https://www.python.org/downloads/ (assuming installation path is added to the path). The script is doing the right thing. (Python from the traditional installer does not have a suffix, it is just python.exe)

e.g. 
image

diablodale:
That is false. searching for `python3` is the right approach.
I challenge you to provide a full directory listing of your python install location.

I just now downloaded python3.9.5 from https://www.python.org/downloads/ and installed it to c:\Python39
Here is the directory listing
```
05/28/2021  01:44 PM    <DIR>          .
05/28/2021  01:44 PM    <DIR>          ..
05/28/2021  01:44 PM    <DIR>          DLLs
05/28/2021  01:44 PM    <DIR>          Doc
05/28/2021  01:44 PM    <DIR>          include
05/28/2021  01:44 PM    <DIR>          Lib
05/28/2021  01:44 PM    <DIR>          libs
05/03/2021  05:35 PM            32,628 LICENSE.txt
05/03/2021  05:37 PM         1,062,280 NEWS.txt
05/03/2021  05:35 PM           101,552 python.exe
05/03/2021  05:35 PM            59,568 python3.dll
05/03/2021  05:35 PM         4,462,768 python39.dll
05/03/2021  05:35 PM           100,016 pythonw.exe
05/28/2021  01:44 PM    <DIR>          Scripts
05/28/2021  01:44 PM    <DIR>          tcl
05/28/2021  01:44 PM    <DIR>          Tools
05/03/2021  05:35 PM            96,120 vcruntime140.dll
05/03/2021  05:35 PM            36,728 vcruntime140_1.dll
               8 File(s)      5,951,660 bytes
              10 Dir(s)  552,720,588,800 bytes free
```

python3 is there, and waiting to be used in the script's search.

diablodale:
Found more info. I think we are both errant, but can likely fix that.

Python and python3 app exec aliases were added to Windows 10 1903 "...every install of Windows will include python and python3 commands that take you directly to the Python store page." https://devblogs.microsoft.com/python/python-in-the-windows-10-may-2019-update/

Above means that these Win 1903 app exec aliases can easily have path priority over any other install of python. And therefore encounter the bug my fix addresses. The buggy method will find the 0 byte exe alias and fail.


https://docs.python.org/3/using/windows.html provides many install methods. So many, that its likely not worth investigating them all. Instead, to take an approach like cmake --> look on path and frequently used install locations, for known exe's.

* python.org download 3.1.4 doesn't have python3.exe.
* python.org download 3.8.0 doesn't have python3.exe. Yet, the 3.8 package from Windows Store *does* have python3.exe. Therefore, the existence of python3 isn't consistent across Windows install methods.
* python.org download 3.9.0 doesn't have python3.exe.
 
python.org download added python3.exe sometime between 3.9.0 and 3.9.5

I recommend the script then
1. check first for python3.exe. This will work for the app exe aliases added in Windows 1903, and python.org downloads 3.9ish and later
2. Fallback to check for python.exe and examine its version stamp.

How about that approach?

PR #1964, opened 2023-12-15
Bugfix/image buffer release

### LINKED ISSUE #1963: Creating an Image in python from a numpy array doesn't work
The function `create_from_ndarray()` in `src/python/k4a/src/k4a/_bindings/image.py` has a number of errors:

- the naming of the modules used (`_nd` instead of `_np`, and `ctypes` instead of `_ctypes`)
- when the defaults for the size of the image are used (all set to 0), the correct values are not fetched from the array.
- `None` is not permitted as the callback in k4a_image_create_from_buffer(), it leads to the following error:

	`ctypes.ArgumentError: argument 7: <class 'TypeError'>: expected CFunctionType instance instead of NoneType`
	
- the reference count on the numpy array passed in should have its reference count incremented so that the array will not be 
  deallocated when the caller deletes it
  
 Related to the callback, in src/python/k4a/src/k4a/_bindings/k4a.py:

- the declaration of the k4a_image_create_from_buffer has the incorrect callback function type (`_memory_allocate_cb` instead of the correct `_memory_destroy_cb`).

icaven:
- names of modules is corrected
- default values for the image size is taken from the given array
- the callback is correctly specified
- the reference count on the numpy array is incremented
- the declaration of `k4a_image_create_from_buffer` has been corrected

### I tested changes on: 
- [ ] Windows
- [x] Linux

### DIFF
--- a/src/python/k4a/setup.py
+++ b/src/python/k4a/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='k4a',
-    version='0.0.2',
+    version='0.0.3',
     author='Jonathan Santos',
     author_email='jonsanto@microsoft.com',
     description='Python interface to Azure Kinect API.',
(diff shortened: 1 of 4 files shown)


icaven:
@microsoft-github-policy-service agree company="Felix & Paul Studios Inc."

